PR #28, opened 2026-02-04
fix: replace --author flag with GIT_AUTHOR env vars in auto-merge

atoomic:
--author is only valid for git commit, not git merge. All non-squash
merges were failing with "unknown option author". Fix: use
GIT_AUTHOR_NAME/EMAIL + GIT_COMMITTER_NAME/EMAIL env vars which work
for all git commands. Also adds env parameter to run_git().

4 new tests (get_author_env + env verification for merge/squash).

### DIFF
--- a/koan/app/git_auto_merge.py
+++ b/koan/app/git_auto_merge.py
@@ -28,15 +28,26 @@ from app.utils import load_config, get_auto_merge_config
 # Low-level git helpers (stateless)
 # ---------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


sukria:
## Review: Ready to merge (once un-drafted)

Important bugfix: `--author` is only valid for `git commit`, not `git merge`. This was causing all non-squash merges to fail.

The fix (using `GIT_AUTHOR_NAME/EMAIL` environment variables) is correct and works for all git commands.

**Status**: Draft PR. Once marked ready, I can merge.

PR #45, opened 2026-02-04
feat: add local LLM support via CLI provider

atoomic:
## Summary

- New `LocalLLMProvider` in `cli_provider.py` — third CLI provider alongside Claude and Copilot
- New `local_llm_runner.py` — standalone agentic loop with built-in tool use (read/write/edit/glob/grep/shell) via OpenAI function calling protocol
- Connects to any OpenAI-compatible API server: **Ollama**, **llama.cpp**, **LM Studio**, **vLLM**
- Configuration via `config.yaml` (`local_llm:` section) and env vars (`KOAN_LOCAL_LLM_*`)
- Documented example models: GLM 4, Kimi K2, Nemotron, Qwen Coder, Code Llama

## Design

The local provider doesn't invoke an external CLI binary — it runs `local_llm_runner.py` as a Python subprocess. The runner handles the agentic loop internally:
1. Sends prompt + system context to the LLM via `/v1/chat/completions`
2. Parses tool_calls from the response (OpenAI function calling format)
3. Executes tools locally and feeds results back
4. Repeats until final text response or max_turns

This integrates seamlessly with existing `subprocess.run(cmd)` patterns across the codebase.

## Configuration

```yaml
cli_provider: "local"

local_llm:
  base_url: "http://localhost:11434/v1"  # Ollama default
  model: "glm4"
  api_key: ""  # Usually empty for local servers
```

## Test plan

- [x] 37 new tests for `local_llm_runner.py` (tool execution, agentic loop, CLI entry point)
- [x] 37 new tests for `LocalLLMProvider` in `test_cli_provider.py` (flag generation, config resolution, provider registry)
- [x] Full suite: 1113 tests pass (was 1039)

### DIFF
--- a/instance.example/config.yaml
+++ b/instance.example/config.yaml
@@ -3,6 +3,32 @@
 # IMPORTANT: Values starting with "CHANGE:" must be updated before running.
 # Kōan will work with sensible defaults for most other settings.
 
+# CLI provider — which AI agent backend to use
+# Options: "claude" (default), "copilot" (GitHub Copilot CLI), "local" (local LLM)
+# Override via KOAN_CLI_PROVIDER env var
+cli_provider: "claude"
+
+# Local LLM configuration (only used when cli_provider: "local")
+# Connects to any OpenAI-compatible API server:
+#   - Ollama:       http://localhost:11434/v1  (default, `ollama serve`)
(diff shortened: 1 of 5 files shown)


atoomic:
While rebasing that change we need to take into account the recently merged provider abstraction which adds support for Copilot. 
Can we also consider a provider for `ollama`?

PR #53, opened 2026-02-05
feat: add copilot as alternative CLI provider

atoomic:
## Summary

- New `cli_provider.py` with `ClaudeProvider` and `CopilotProvider` classes for CLI backend abstraction
- All 12 CLI invocation sites (7 Python modules + run.sh) use `build_full_command()` for provider-aware command generation
- Tool name mapping: Bash→shell, Read→read_file, Write→write_file, Edit→edit_file, Glob→glob, Grep→grep
- Config: `cli_provider` in config.yaml + `KOAN_CLI_PROVIDER` env var override

Isolated from PR #45 as a standalone feature. This is the first of 3 commits being split out.

## Test plan

- [x] 57 new tests in `test_cli_provider.py` (both providers + helpers)
- [x] Full suite: 900 tests pass on upstream/main base

### DIFF
--- a/instance.example/config.yaml
+++ b/instance.example/config.yaml
@@ -55,6 +55,11 @@ tools:
     - Search: Glob (find files), Grep (search content)
     - Shell: Bash (run git, gh, make, and other commands)
 
+# CLI provider — which AI CLI to use as the backend
+# Options: "claude" (default), "copilot" (GitHub Copilot CLI)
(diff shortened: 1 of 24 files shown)


atoomic:
Make sure no new code from upstream are introducing specific logic to `claude` we need to prefer using the abstraction provider layer we introduced here

PR #56, opened 2026-02-05
fix(awake): handle CTRL-C gracefully without traceback

atoomic:
## Summary
- Wrap the `main()` polling loop with `try/except KeyboardInterrupt` so CTRL-C prints `[awake] Shutting down.` and exits cleanly instead of dumping a 30-line Python backtrace through requests/urllib3/ssl
- Worker threads are already daemonic — no cleanup needed

## Test plan
- [x] 2 new tests: `test_main_ctrl_c_exits_gracefully` and `test_main_ctrl_c_during_sleep_exits_gracefully`
- [x] Full suite: 802 tests pass

### DIFF
--- a/koan/app/awake.py
+++ b/koan/app/awake.py
@@ -876,20 +876,24 @@ def main():
     print(f"[awake] Polling every {POLL_INTERVAL}s (chat mode: fast reply)")
     offset = None
 
-    while True:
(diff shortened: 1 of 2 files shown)


atoomic:
Confirmed that if fixes the issue, this is now ready to review/merge @sukria

Issue #970, opened 2026-03-21
Rebase should detect meaningless remains.


toddr:
If when rebasing, you determine that the diff from main that is left is meaningless (comments, meta data) and no meaningful change, then consider determining what PR or commit fixed the work and then close the pull request automatically, referencing that this is fixed by XXX already. We should also close the referred to issue if there was one when opening the PR since that issue is now fixed by XXX and we can note it in the issue and close or at least comment in it.

PR #983, opened 2026-03-21
feat: auto-close rebased PRs with only meaningless remains

mateu:
## What
Auto-close rebased PRs when the remaining diff is non-functional (comments/docs/metadata only).

## Why
Issue #970 asks rebase to detect when a PR has no meaningful remaining work and close it as already addressed instead of recycling stale noise.

## How
- Added remaining-diff analysis in `rebase_pr` with a conservative meaningful-change heuristic.
- Added superseding reference lookup from base history and PR auto-close comment (`fixed by ...`).
- Added linked-issue follow-up (comment + best-effort close) when issues are referenced in the PR body.
- Guarded auto-close to only trigger when there is no review-feedback context.

## Testing
- `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_rebase_pr.py -q`
- `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_rebase_pr.py koan/tests/test_silent_exceptions.py -q`
- `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests -q` *(baseline has unrelated failures in format_outbox/reset_parser)*

### DIFF
--- a/koan/tests/test_rebase_pr.py
+++ b/koan/tests/test_rebase_pr.py
@@ -15,6 +15,8 @@ from app.rebase_pr import (
     fetch_pr_context,
     build_comment_summary,
     run_rebase,
+    _diff_has_meaningful_changes,
+    _extract_linked_issue_numbers,
     _apply_review_feedback,
     _build_ci_fix_prompt,
     _build_rebase_comment,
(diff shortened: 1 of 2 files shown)


koan/tests/test_rebase_pr.py
```diff
@@ -468,6 +470,57 @@ def test_partial_comments(self):
...
+            "diff --git a/app.py b/app.py\n"
+            "--- a/app.py\n"
+            "+++ b/app.py\n"
+            "@@ -1 +1 @@\n"
+            "-# old note\n"
+            "+# updated note\n"
+        )
+        assert _diff_has_meaningful_changes(diff, ["app.py"]) is False
+
+    def test_docs_only_change_is_not_meaningful(self):
```
toddr: unfortunately "Meaningful" can be a problematic word. I can see a README update being important. :( 

I would be tempted to say we should be solving this by changing the LLM template to decide if the rebase it gave back is meaningful in solving the issue rather than these static rules.

mateu:
> I do not think this is going to work to try to close issues based on files changes

I agree, ok if if close this PR?

Issue #1042, opened 2026-03-26
Add 'mode' option: active (default) vs passive


sukria:
## Summary

Koan should support two operating modes: **active** (default) and **passive**.

### Active (default)
Current behavior — Koan takes over missions, goes into deep research, switches branches, etc.

### Passive
Koan is **strictly read-only**. The agent loop keeps running (heartbeat, GitHub notification polling, Telegram commands) but **never executes missions or autonomous work**. No Claude CLI calls, no branch switching, no code writing, no commits, no PRs.

Key constraint: **passive mode must never modify the local git clone in any way**. Passive == the human is working on this machine.

Missions from GitHub notifications or Telegram still get queued to `missions.md` as Pending — they drain when active mode resumes.

## How it differs from existing mechanisms

| Mechanism | Loop runs? | Missions execute? | Exploration? | GitHub poll? | Telegram? |
|-----------|-----------|-------------------|-------------|-------------|-----------|
| **Pause** | Blocked in sleep loop | No | No | No | Yes (commands) |
| **Focus** | Yes | Yes (missions only) | No | Yes | Yes |
| **Work hours** | Yes | Yes | Suppressed | Yes | Yes |
| **Passive** | Yes (idle) | **No** | **No** | **Yes (queue only)** | **Yes** |

## New Commands

| Command | Behavior |
|---------|----------|
| `/passive [duration]` | Force passive mode immediately. Optional duration (e.g., `/passive 4h`). No duration = indefinite. |
| `/active` | Resume active mode immediately |

## Update existing commands

| Command | Change |
|---------|--------|
| `/status` | Show current mode: `👁️ Passive (read-only)` or `🟢 Active` |

## Configuration

`start_passive: true` in `config.yaml` — start in passive mode at boot (for devs who always want Koan inert by default).

## Motivation

When a developer is actively working on the same machine, Koan's autonomous behaviors (branch switching, deep research, mission execution) can be disruptive. Passive mode lets Koan stay alive (queuing notifications, responding to commands) without interfering with the human's workflow.

## Implementation Plan

See [revised plan (iteration 2)](https://github.com/Anantys-oss/koan/issues/1042#issuecomment-4137259487) for full implementation details.

sukria:
Fixed in main. Closing.

PR #1053, opened 2026-03-27
feat: add passive/active mode toggle

sukria-koan0:
## Summary

Add a passive/active mode toggle for Kōan. Passive mode keeps the agent loop running (heartbeat, GitHub notification polling, Telegram commands) but blocks all execution — no missions, no exploration, no Claude CLI calls. Missions accumulate as Pending and drain naturally when `/active` is used.

Closes https://github.com/Anantys-oss/koan/issues/1042

## Changes

- **`passive_manager.py`**: New state manager following `focus_manager.py` pattern — `PassiveState` dataclass with indefinite + timed duration, auto-expiry, atomic writes
- **`signals.py`**: Add `PASSIVE_FILE = ".koan-passive"` constant
- **`iteration_manager.py`**: Gate execution at step 4b — passive check before `start_mission()`, returns `passive_wait` action
- **`run.py`**: Add `passive_wait` to `_IDLE_WAIT_CONFIG` with status display
- **`/passive` and `/active` skills**: Telegram commands for toggling mode (`/passive [duration]`, `/active`)
- **`/status`**: Shows passive/active mode label
- **`start_passive` config**: Boot into passive mode when `start_passive: true` in `config.yaml`
- **50 new tests** covering manager, skill handler, status integration, and iteration_manager integration

## Test plan

- [x] `test_passive_manager.py` — 34 unit tests for state lifecycle, expiry, duration display
- [x] `test_passive_integration.py` — 6 integration tests verifying iteration_manager returns `passive_wait`
- [x] `test_passive_manager.py` — 10 skill handler + status integration tests
- [x] Full suite: 10918 tests passing, no regressions

---
*Generated by Kōan /implement*

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -77,6 +77,7 @@ Communication between processes happens through shared files in `instance/` with
 - **`pause_manager.py`** — Pause state management (`.koan-pause` / `.koan-pause-reason` files). Supports time-bounded pauses with auto-resume (e.g., `/pause 2h`)
 - **`restart_manager.py`** — File-based restart signaling between bridge and run loop (`.koan-restart`)
 - **`focus_manager.py`** — Focus mode management (`.koan-focus` JSON); skips contemplative sessions when active
+- **`passive_manager.py`** — Passive mode management (`.koan-passive` JSON); read-only mode that blocks all execution while keeping loop alive
 
 **CLI provider abstraction** (`koan/app/provider/`):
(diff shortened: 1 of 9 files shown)


atoomic:
@sukria-koan0 ask does the `/status` output shows if we are in passive mode? it s important for the user to know if we are in passive mode. We should also display the time remaining if a timer was used
No need to advertise the active mode, should be the default, unless it s for a limited period of time.

PR #1176, opened 2026-04-10
feat: add strict_missions mode to gate autonomous work (#1175)

### LINKED ISSUE #1175: Add strict_missions mode: disable all autonomous work, only run user-queued missions
## Problem

Some users want Kōan to run as a pure "mission executor" — a quiet bot that only acts when explicitly told to. Today, even with `exploration: false` set per-project, Kōan still:

- Enters **DEEP mode** when budget is ample (`koan/app/usage_tracker.py` `decide_mode()`), which encourages the agent to pick up GitHub issues autonomously (per `koan/system-prompts/agent.md` "GitHub Issue Selection").
- Runs **contemplative sessions** on random rolls in deep/implement modes (`koan/app/iteration_manager.py` `_should_contemplate()`).
- Falls back to **autonomous exploration** when no mission is pending (`koan/app/iteration_manager.py` `_decide_autonomous_action()`).

Setting `exploration: false` on every project only blocks fallback #3 — contemplative sessions still fire, and the agent prompt still encourages autonomous GitHub issue picking when in DEEP mode.

`passive` mode is too strict (it blocks mission execution entirely), and quota-based gating is orthogonal.

## Requested behavior

A single global switch — proposed `strict_missions: true` in `config.yaml` (env override: `KOAN_STRICT_MISSIONS=1`) — that enforces all of the following:

1. **Only explicit missions run.** Missions come from two sources:
   - `/mission` sent via Telegram (already queues to `missions.md`).
   - GitHub notifications: @mention-triggered missions via `koan/app/github_command_handler.py` continue to work.
2. **No autonomous exploration.** When `missions.md` has no Pending items, the loop idles (like `exploration_wait`) instead of calling `_select_random_exploration_project()`.
3. **No contemplative sessions.** `_should_contemplate()` returns False unconditionally.
4. **No DEEP mode.** `UsageTracker.decide_mode()` (or a post-decision cap in `iteration_manager._get_usage_decision`) caps at `implement` max. Prevents the agent prompt from being told "you have ample budget, pick an issue".
5. **No autonomous GitHub issue pickup.** The agent prompt (`koan/system-prompts/agent.md`) must be adjusted so that under `strict_missions`, the "GitHub Issue Selection" section is disabled — the agent only acts on issues explicitly referenced by a queued mission.
6. **Recurring missions still fire.** They're user-authored, just scheduled — keep them working.
7. **Auto-update, heartbeat, Telegram polling, GitHub notification polling, CI queue drain** all continue to run — this mode gates *work selection*, not the bridge.

## Suggested implementation sketch

- **Config:** Add `strict_missions: false` (default) to `instance.example/config.yaml`, with a helper `is_strict_missions()` in `koan/app/config.py` reading `KOAN_STRICT_MISSIONS` env var first, then `config.yaml`.
- **Mode cap:** In `iteration_manager.py` after `decide_mode()`, cap `autonomous_mode = "implement"` if `deep` when strict is on (or introduce a new `_cap_mode_for_strict()` helper mirroring `cap_mode_for_schedule`).
- **Exploration short-circuit:** In `iteration_manager.py` the "No mission — autonomous mode" branch, when `strict_missions` is on, return a new `action="strict_wait"` result immediately, before `_filter_exploration_projects()`.
- **Contemplative gate:** In `_should_contemplate()`, return False when strict is on.
- **Prompt:** Pass a `strict_missions` flag into `koan/app/prompt_builder.py` and conditionally omit or replace the "GitHub Issue Selection" section of the agent prompt.
- **Loop manager:** `koan/app/loop_manager.py` should treat `strict_wait` like `exploration_wait` for sleep/wake-on-mission behavior, so a new `/mission` wakes it immediately.
- **Tests:**
  - `plan_iteration` returns `action="strict_wait"` when no missions are pending.
  - `decide_mode()` never returns `"deep"` under strict.
  - `_should_contemplate()` returns False under strict even with ample budget.
  - A queued mission still runs normally under strict.
  - A GitHub @mention-triggered mission is still accepted.
  - Recurring injection still works.
- **Docs:** Update `docs/user-manual.md` and `README.md` to describe the mode and contrast it with `passive` (passive = no execution at all; strict_missions = no autonomous work but missions still execute).

## Out of scope

- Removing or weakening existing `exploration:` per-project flag (kept as finer-grained override).
- Changing `passive` semantics.
- A Telegram `/strict` toggle — can come later; v1 is config-only.

sukria-koan0:
## Summary

Closes #1175.

Adds a single global switch — `strict_missions` — that turns Kōan into a pure mission executor. When enabled, only missions that were explicitly queued (via `/mission`, recurring schedules, or GitHub @mention commands) ever run. All autonomous work is gated off.

## What this changes

- **New config helper** `is_strict_missions()` in `koan/app/config.py`, resolving from `KOAN_STRICT_MISSIONS` env var first, then `strict_missions:` in `config.yaml` (default: `false`).
- **Mode cap** — `plan_iteration()` caps `decide_mode()` at `implement` when strict is on, so the agent prompt never gets the "ample budget, pick an issue" framing of DEEP mode.
- **`strict_wait` action** — when no mission is pending, `plan_iteration()` short-circuits to a new `strict_wait` result *before* exploration filtering, skipping `_filter_exploration_projects()` entirely (avoids wasted `gh` calls). It idles with wake-on-mission like `exploration_wait`.
- **Contemplative gate** — `_should_contemplate()` unconditionally returns `False` under strict mode.
- **Agent prompt override** — `prompt_builder._apply_strict_missions_override()` replaces the `## GitHub Issue Selection` section with a `## Strict Missions Mode` block that explicitly forbids autonomous GitHub pickup.
- **Run loop** — `run.py` handles `strict_wait` via the idle-wait config map with a clear status message.
- **Recurring, Telegram polling, GitHub notification polling, CI queue drain, auto-update, heartbeat** — all continue to run. Strict mode gates *work selection*, not the bridge.

## What it is NOT

- Not a replacement for `/passive` — passive blocks all execution (missions sit as Pending). Strict mode keeps running queued missions.
- Not a weakening of the per-project `exploration:` flag — that's still the finer-grained override.
- No Telegram `/strict` toggle yet — v1 is config-only, as specified in the issue.

## Tests

Added 14 new tests in `koan/tests/test_iteration_manager.py`:

- `TestStrictMissionsContemplate` — `_should_contemplate()` returns False under strict even with ample budget / deep mode / implement mode.
- `TestStrictMissionsPlanIteration` — covers:
  - no-mission → `strict_wait` (DEEP capped to implement)
  - exploration filter is never called
  - queued missions still run
  - contemplative rolls are blocked
  - recurring injection still fires
- `TestStrictMissionsConfigHelper` — env-var truthy, env-var false overrides yaml, yaml fallback, default false.
- `TestStrictMissionsPromptOverride` — GitHub section replaced when strict; intact when not.

Full suite (**11,530 tests**) passes locally.

## Docs

- `instance.example/config.yaml` — new commented `strict_missions: false` block explaining the option and env override, and contrasting it with passive mode.
- `docs/user-manual.md` — new "Strict Missions Mode" section under the passive mode docs with enable/disable instructions, what continues vs what's gated, and when to use it.

## Test plan

- [x] Unit tests pass (`KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_iteration_manager.py`)
- [x] Full suite passes (`KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/`)
- [ ] Manual smoke: set `strict_missions: true`, start run loop, verify `/status` reports `strict_wait`, queue a mission via `/mission`, verify it runs, remove config, verify normal autonomous behavior returns.

### DIFF
--- a/docs/user-manual.md
+++ b/docs/user-manual.md
@@ -279,6 +279,37 @@ Kōan can manage multiple projects simultaneously. It rotates between them based
 - `/active` — "I'm done, you can work again"
 </details>
 
+### Strict Missions Mode
+
+**Strict missions** is a config-level switch that turns Kōan into a pure mission executor. When enabled, the agent only runs missions you explicitly queue — it never picks up GitHub issues autonomously, never runs contemplative reflection, and never enters DEEP mode. The loop keeps polling Telegram, GitHub notifications, and recurring schedules, so it still wakes up the moment you queue something.
+
+- **Enable in `instance/config.yaml`:**
+  ```yaml
+  strict_missions: true
+  ```
+- **Or via environment variable:** `KOAN_STRICT_MISSIONS=1` (takes precedence over `config.yaml`).
+- **Disable:** set back to `false`, or `KOAN_STRICT_MISSIONS=0`.
+
+What continues to run under strict mode:
+
+- Missions queued via `/mission`, GitHub `@mention` commands, and recurring schedules.
+- Heartbeat, auto-update, Telegram polling, GitHub notification polling, CI queue drain.
+
+What is disabled:
+
+- DEEP mode (capped at `implement`).
+- Contemplative sessions (random reflection rolls are skipped).
+- Autonomous exploration (`strict_wait` replaces `exploration_wait` when no mission is pending).
+- The agent prompt's `GitHub Issue Selection` section is replaced with an explicit "do not pick up issues" instruction.
+
+**How it differs from `/passive`:** passive mode blocks all execution (missions sit as Pending until you `/active`). Strict missions mode keeps the executor running for any mission you queue — it only gates *autonomous work selection*.
+
+**When to use:**
+
+- You want Kōan to act strictly on demand, no surprises on the PR list.
+- You're handing off mission dispatch to another system (CI, a team workflow) and want Kōan to be a quiet executor.
+- Multi-bot setups where only one instance should pick up issues autonomously.
+
 ---
 
 ## Intermediate — Productivity Workflows
--- a/docs/user-manual.md
+++ b/docs/user-manual.md
@@ -279,18 +279,28 @@ Kōan can manage multiple projects simultaneously. It rotates between them based
 - `/active` — "I'm done, you can work again"
 </details>
 
-### Strict Missions Mode
+### Permanent Focus Mode
 
-**Strict missions** is a config-level switch that turns Kōan into a pure mission executor. When enabled, the agent only runs missions you explicitly queue — it never picks up GitHub issues autonomously, never runs contemplative reflection, and never enters DEEP mode. The loop keeps polling Telegram, GitHub notifications, and recurring schedules, so it still wakes up the moment you queue something.
+Focus mode can be made permanent via config, turning Kōan into a pure mission executor. When enabled, the agent only runs missions you explicitly queue — it never picks up GitHub issues autonomously, never runs contemplative reflection, and never enters DEEP mode. The loop keeps polling Telegram, GitHub notifications, and recurring schedules, so it still wakes up the moment you queue something.
 
-- **Enable in `instance/config.yaml`:**
+This extends the `/focus` Telegram command (which is time-bounded) into a permanent config-level switch.
+
+- **Enable globally in `instance/config.yaml`:**
   ```yaml
-  strict_missions: true
+  focus: true
   ```
-- **Or via environment variable:** `KOAN_STRICT_MISSIONS=1` (takes precedence over `config.yaml`).
-- **Disable:** set back to `false`, or `KOAN_STRICT_MISSIONS=0`.
+- **Or via environment variable:** `KOAN_FOCUS=1` (takes precedence over `config.yaml`).
+- **Per-project in `projects.yaml`:**
+  ```yaml
+  defaults:
+    focus: true          # All projects focused by default
+  projects:
+    myapp:
+      focus: false       # Override: allow autonomous work on myapp
+  ```
+- **Disable:** set back to `false`, or `KOAN_FOCUS=0`.
 
-What continues to run under strict mode:
+What continues to run under focus mode:
 
 - Missions queued via `/mission`, GitHub `@mention` commands, and recurring schedules.
 - Heartbeat, auto-update, Telegram polling, GitHub notification polling, CI queue drain.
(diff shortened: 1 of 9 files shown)


sukria:
could be good to add the `/strict` Telegram hook.

atoomic:
Would need /strict to turn it on and /unstrict to disable it from telegram

atoomic:
@sukria-koan0 rebasw

atoomic (approved):

atoomic:
Note that we have the /focus mode

> 💡 Focus mode is live. Use /focus to suppress reflection and free exploration and just process the mission queue. Default is 5 hours. You can dial it: /focus 3h, /focus 2h30m, whatever you need. Keep it lean.

sukria:
Good point @atoomic , thanks for pointing it out. The user wants a full permanent mission picker though. With Github polling.
I think that strict mode deserves to exist ;)

atoomic:
I think the `/strict` mode is the same as the `/focus` mode
We should only keep one single name, and this is great to have both the config variable and the ability to permanently disable / enable it from telegram while keeping the ability to temporary disable it for a few hours so we do not have to think to bring it back online

sukria:
Agreed, let's extend /focus to support the workflow described here.

Also, we should be able to enable or disable this focus mode on a per project basis.

Config should allow a default (applies to all projects) and per project overwrite.

@sukria-koan0 rebase

sukria-koan0:
## Rebase: feat: add strict_missions mode to gate autonomous work (#1175)

Branch `koan/1175-strict-missions-mode` rebased onto `main` and force-pushed.

**Diff**: 9 files changed, 498 insertions(+), 8 deletions(-)

Review feedback was analyzed and applied.

### Changes

builder.py` — `_is_strict_missions()` → `_is_focus_mode()`, `_apply_strict_missions_override()` → `_apply_focus_mode_override()`, replacement text says "Focus Mode" instead of "Strict Missions Mode". Also enhanced `_is_focus_mode()` to check both config and `.koan-focus` file
- **Removed `strict_wait`** from `koan/app/run.py` idle-wait config map; updated `focus_wait` default display to say "permanent" instead of "unknown"
- **Updated `instance.example/config.yaml`** — replaced `strict_missions` block with `focus` block including per-project override docs
- **Updated `docs/user-manual.md`** — replaced "Strict Missions Mode" section with "Permanent Focus Mode" including per-project `projects.yaml` examples
- **Updated `projects.example.yaml`** — added `focus` config documentation in defaults section and a per-project example
- **Updated all 14 tests** in `koan/tests/test_iteration_manager.py` — renamed classes, methods, mock targets, and assertions from `strict_missions` to `focus_mode`

### Actions

- Already-solved check: negative (confidence=high, reasoning=None of the recent commits on main implement a strict_missions mode or anything semantically equival)
- Rebased `koan/1175-strict-missions-mode` onto `origin/main`
- Applied review feedback
- Force-pushed `koan/1175-strict-missions-mode` to origin

### CI

---
_Automated by Kōan_

sukria:
Looks very good indeed. With user manual and sample config entries, per-project overrides, nice job Koan!

PR #1191, opened 2026-04-14
fix: load .env early in run.py so env vars reach Claude subprocesses

herbertia:
Custom env vars like NOTION_API_KEY and SENTRY_ACCESS_TOKEN were not
available to Claude CLI subprocesses because load_dotenv() was called
too late. Move it to main_loop() startup before any subprocess is spawned.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/koan/app/run.py
+++ b/koan/app/run.py
@@ -698,6 +698,11 @@ def main_loop():
         log("error", "No instance/ directory found. Run: cp -r instance.example instance")
         sys.exit(1)
 
+    # Load .env so custom env vars (NOTION_API_KEY, SENTRY_ACCESS_TOKEN, etc.)
+    # are available to Claude subprocesses that inherit os.environ.
+    from app.utils import load_dotenv
+    load_dotenv()
+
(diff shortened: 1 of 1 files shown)


koan/app/run.py
```diff
@@ -698,6 +698,11 @@ def main_loop():
         log("error", "No instance/ directory found. Run: cp -r instance.example instance")
         sys.exit(1)
 
+    # Load .env so custom env vars (NOTION_API_KEY, SENTRY_ACCESS_TOKEN, etc.)
+    # are available to Claude subprocesses that inherit os.environ.
+    from app.utils import load_dotenv
+    load_dotenv()
```
atoomic, line 704: I do not think it s a good idea to load and expose all ENV variables to the cli 
If a custom skill need a custom env variable, it should read it from another custom file, or even better
`~/.notion_api_key` so it s very targeted

atoomic (changes requested):
Not a good idea

PR #1266, opened 2026-04-28
refactor: centralize atomic-write pattern via utils.atomic_write_json

atoomic:
The mkstemp + fsync + os.replace + cleanup-on-error pattern was duplicated across four modules, with subtle drift (attention.py used NamedTemporaryFile without fsync; conversation_history.py kept a local copy explicitly to dodge a circular import).

Add atomic_write_json() to utils.py as a thin JSON wrapper around the existing atomic_write(), and route session_manager, attention, reaction_store, and conversation_history through the centralized helpers via lazy imports. Net -44 lines and one less inconsistent fsync gap.

### DIFF
--- a/koan/app/attention.py
+++ b/koan/app/attention.py
@@ -13,12 +13,9 @@ Severities: critical > warning > info.
 Dismissed items are tracked in instance/.koan-attention-dismissed.json.
 """
 
-import fcntl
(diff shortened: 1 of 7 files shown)


atoomic:
@Koan-Bot review

PR #1327, opened 2026-05-15
feat: forward mission result text to outbox for SKIP/FAIL/ERROR outcomes

atoomic:
Adds _notify_mission_result() to the post-mission pipeline so the Claude session's final result string reaches Telegram even when the session's sandbox blocked writes to instance/outbox.md. Activates when the result contains alert markers (SKIP/FAIL/ERROR/BLOCKED, "permission deadlock", "hard stop", etc.) or when the mission title matches a customer-facing skill (/cpfix, wp-bug-resolver).

Idempotent: skipped silently when the session has already written to outbox.md after the mission start time. Posts at ACTION priority so the message clears the default min_priority filter; the alert flag only toggles the icon (⚠️ for alerts, ℹ️ for customer-facing success).

Gated by the new notify_mission_results config key (default: True) so operators can opt out.

### DIFF
--- a/docs/github-commands.md
+++ b/docs/github-commands.md
@@ -251,7 +251,7 @@ The helper is `app.external_skill_dispatch.try_dispatch_custom_handler`. It also
 
 - **Jira source**: the issue the comment is on.
 - **GitHub source**: the first `FOO-123`-style key found in the issue title, then body.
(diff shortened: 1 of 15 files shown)


atoomic:
@Koan-Bot review

PR #1328, opened 2026-05-15
fix(jira): pick up @mentions ranked deep in result set

atoomic:
The notification poller was silently dropping legitimate @mentions because fetch_jira_mentions() capped results at the 20 most-recently-updated issues across all mapped projects. On a multi-project deployment, 24h of activity routinely produces 50–100 issues, so anything not in the top 20 by `updated DESC` never had its comments inspected.

Empirically, an issue ranked 46 of 100 in production carried a legitimate `@<bot> <command>` mention but the polling cycle logged "No new Jira notifications" and the mention was lost. JQL pre-filtering by mention text isn't viable either: `text ~ "<nickname>"` and `comment ~ ...` both miss ADF mention nodes because Jira indexes the accountId reference, not the displayName.

Changes:
- Bump _MAX_ISSUES_PER_CYCLE from 20 to 200. Cold-start cost rises from 21 to ~201 API calls (~20s at 10 req/s), comparable to GitHub's "~1 min" cold start. Subsequent polls use _last_jira_check_iso (≤60s window) so steady-state cost is unchanged.
- Promote the per-cycle search log from DEBUG to INFO and surface the JQL since-window + result count, so future "no mentions" mysteries are diagnosable from run.log alone.
- When the cap actually clips, emit a WARNING with guidance to tighten max_age_hours or shorten check_interval_seconds.
- Add a Telegram cold-start banner "🔍 Scanning Jira notifications..." parallel to GitHub's existing one, so the user sees that Jira IS being scanned in the same shape and time as GitHub.

CLAUDE.md gains a "Never leak private skill/agent/project names" convention documenting the policy and the pre-commit grep check (this commit also scrubs three pre-existing leaks of private project keys in test fixtures and a docstring).

Regression test: stub a 100-issue result set with the target mention at index 46 and assert fetch_jira_mentions still returns it. With the old cap of 20 this test would fail.


atoomic:
@Koan-Bot review

PR #1330, opened 2026-05-15
fix(github): persist dedup for assignment notifications

atoomic:
review_requested and assign notifications kept re-queueing the same mission after every restart, because none of the existing dedup layers covered the case:

  - The in-memory _notif_cache in loop_manager.py is lost on restart.
  - The persistent comment tracker keys on comment IDs, but these notifications have no comment object to react to or hash.
  - The fallback missions.md check in _try_assignment_notification only scans the Pending section, so once the prior /review moved to In Progress/Done/Failed it became invisible.

Add a parallel persistent tracker in github_notification_tracker.py that records (notification_id, updated_at) composite keys for 7 days in instance/.koan-github-processed-threads.json. Wire it into _try_assignment_notification with an early-return guard above the staleness/closed/repo checks, plus track_thread calls on both the "already pending" and "successfully inserted" paths so subsequent restarts short-circuit cleanly.

The composite key naturally invalidates when GitHub bumps updated_at (re-requested review, new commits pushed), so a renewed request still queues a fresh mission rather than being permanently silenced.


atoomic:
@Koan-Bot review

PR #1332, opened 2026-05-15
feat(hooks): discover skill-bound lifecycle modules

atoomic:
Lets a custom skill own its full mission lifecycle (e.g. enforce a JIRA outcome comment after every fix mission) without touching Kōan core. Previously hooks could only live in instance/hooks/ and were instance-wide; skill authors had no way to ship pre/post-mission logic alongside their handler.py.

HookRegistry now also walks instance/skills/<scope>/<name>/ for files named after a known event (session_start.py, session_end.py, pre_mission.py, post_mission.py) and registers each module's run(ctx) as a handler for that event. Instance-wide hooks still fire first; skill-bound hooks run after. Errors stay isolated per handler.

mission_runner._fire_post_mission_hook now pre-reads the truncated Claude stdout via _read_stdout_summary and passes it to hooks as ctx['result_text'], so hooks can parse JIRA keys / PR URLs / RESULT lines without re-implementing file I/O.

Makefile gains a test-skills target that auto-discovers instance/skills/**/tests/test_*.py (handles symlinked skill repos via find -L) and runs them with KOAN_REPO + PYTHONPATH set up. It chains into make test and make test-strict so skill regressions surface in the regular suite; skipped cleanly when no skill tests exist.

instance.example/hooks/README.md documents both hook flavors, the new result_text ctx key, and the convention for shipping tests alongside a skill (tests/conftest.py template + pytest invocation recipes).

Tested: 224 koan tests pass (52 existing hook + 9 new skill-bound discovery + 163 mission_runner); 6 pre-existing environment failures are unrelated (confirmed via git stash). Tests cover discovery, ordering, error isolation, ctx contents, and the missing-skills-dir edge case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 export
 
 .PHONY: install onboard setup start stop status restart
-.PHONY: clean say migrate test test-strict coverage sync-instance rename-project release
+.PHONY: clean say migrate test test-skills test-strict coverage sync-instance rename-project release
 .PHONY: awake run errand-run errand-awake dashboard
 .PHONY: ollama logs ssh-forward
 .PHONY: install-systemctl-service uninstall-systemctl-service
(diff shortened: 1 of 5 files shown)


atoomic:
@Koan-Bot review

PR #1363, opened 2026-05-17
feat(watchdog): runner-side self-heal for stale or hung bridge

atoomic:
The Telegram bridge (`awake.py`) is a long-running process with no
wrapper, so a frozen `sys.modules` after `/update` leaves it serving
stale imports until someone with shell access kicks it.  The recent
`cb6e927` ("per-process restart markers") fix the race that drops
restart signals once both processes are on the new code, but the
transitional `/update` from old to new still wedged the bridge — and
no in-Telegram recovery path worked, because every subsequent
`/restart` hit the same import-cache state and the bridge's polling
loop has no way to bootstrap its `awake.py` globals back to the
freshly-reloaded `app.restart_manager`.

Add a runner-side watchdog that detects two distinct failure modes
and escalates recovery through four tiers.  The runner is *always*
fresh code (its wrapper relaunches on every exit), so it is the
right place to own this responsibility.

Detection:
- **Stale modules**: bridge stamps its git HEAD SHA at startup
  (`.koan-bridge-version`); runner compares against
  `git rev-parse HEAD` on each tick.  Heartbeat alone cannot catch
  this — process is alive and responsive, just running old code.
- **Hung / dead bridge**: heartbeat older than
  `BRIDGE_HEARTBEAT_STALE_S` (90 s) or PID gone / pointing at a
  dead process.

Escalation (one tier per call, persisted in `.koan-bridge-heal-state`):
- Tier 1: `request_restart()` — runner is on fresh code so its
  triple-marker write is correct and not subject to the old race
- Tier 2: SIGTERM the bridge PID
- Tier 3: SIGKILL (after `SIGTERM_GRACE_S`) + `pid_manager.start_awake`
- Tier 4: circuit-broken, alert and stop.  Counter trips after
  `HEAL_CIRCUIT_BREAKER_LIMIT` (3) consecutive tier-3 failures.

If the bridge has no PID file at all (crashed long ago, never
restarted), tiers 1–2 are skipped and the watchdog goes straight to
tier 3 — there is no process to signal cooperatively.

The runner throttles probes to one per `_BRIDGE_WATCHDOG_INTERVAL`
(5) main-loop iterations.  Each iteration is ~60–300 s, so detection
latency is ~5–25 min in the worst case, which is fine for "don't let
a stuck bridge sit forever."  Per-tier cooldown
(`HEAL_TIER_COOLDOWN_S` = 45 s) prevents acting again before the
previous tier has had a chance to take effect.

When the watchdog acts, the message is forwarded both to Telegram
via `_notify_raw` (terse, no Claude reformat — goes through the
outbox so a freshly-relaunched bridge flushes it on first poll) and
to today's journal entry under project "koan" for after-the-fact
audit.

Files:
- `koan/app/bridge_watchdog.py` (new, ~330 LOC): detection, state
  machine, tier execution, bridge-side version stamp writer.
- `koan/app/signals.py`: two new constants
  (`BRIDGE_VERSION_FILE`, `BRIDGE_HEAL_STATE_FILE`).
- `koan/app/awake.py`: call `write_bridge_version_stamp` once at
  startup, right after the existing heartbeat write.
- `koan/app/run.py`: throttled `_maybe_run_bridge_watchdog` helper
  called once per iteration after the existing `check_restart`
  block.  Forwards heal messages to Telegram + journal.
- `koan/tests/test_bridge_watchdog.py` (12 tests): healthy no-op,
  SHA mismatch tier 1, heartbeat stale tier 1, missing stamp =
  no-op (defensive), cooldown suppression, tier 2 SIGTERM after
  tier 1, tier 3 SIGKILL + start_awake, no-PID cold-start
  skipping straight to tier 3, circuit-breaker alert path.
- `docs/bridge-watchdog.md`: design rationale, failure-mode table,
  escalation diagram, tunables, operator notes, rollback.

Verified: `pytest koan/tests/` = 12788 passed, ruff clean on all
touched files.  (Pre-existing SIM105 in `koan/app/cli_exec.py:187`
is unrelated and present on `main`.)

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/docs/bridge-watchdog.md
+++ b/docs/bridge-watchdog.md
@@ -1,0 +1,178 @@
+# Bridge Self-Heal Watchdog
+
+The Telegram bridge (`awake.py`) is a long-running process **with no
+wrapper**: once started, nothing restarts it automatically. If its
+`sys.modules` cache goes stale (after `/update` pulls new code), the
+bridge keeps serving the *old* code until an operator with shell access
+kills it. The original `cb6e927` ("per-process restart markers") fix
+patched the race that drops restart signals, but only *after* both
+processes are running the new code — the transitional `/update` from
+old to new can still leave the bridge wedged.
+
+The watchdog runs **inside the agent loop** (`run.py`) and recovers a
+stale or hung bridge automatically.
+
+## Why this lives in the runner, not the bridge
+
+`run.py` is wrapped by `run.sh` and restarted on every exit. Whatever
+SHA is on disk is what the runner is executing — it can't be stale.
+That makes it the natural place to watch the bridge.
+
+## Two failure modes detected
+
+| Failure mode | How it shows up | How it's detected |
+|---|---|---|
+| **Stale `sys.modules`** | Bridge is alive, heartbeat fresh, but `/list` and other skills raise `ImportError` on names added by the update. | Stamp the bridge's git HEAD at startup; runner compares with `git rev-parse HEAD` on each tick. |
+| **Hung / dead bridge** | Heartbeat mtime stops advancing; possibly the process is gone entirely. | Read `.koan-heartbeat` mtime; read `.koan-pid-awake` and probe with `kill(pid, 0)`. |
+
+## Four-tier escalation
+
+```
+                                       ┌─ healthy → reset state, return None
+            ┌─── unhealthy? ─── yes ───┤
+unhealthy = │                          └─ in cooldown? ── yes ─── return None
+  sha drift │                                      no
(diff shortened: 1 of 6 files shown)


atoomic:
@Koan-Bot review

atoomic:
@Koan-Bot rebase

atoomic:
@Koan-Bot review

atoomic:
Probably overkill regarding the value. All the most the Makefile and systemd services need some adjustments

PR #1422, opened 2026-05-21
fix(notifications): don't mark foreign-repo notifications as read

atoomic:
When multiple Kōan instances share a single GitHub bot identity, each
    instance must only touch notifications for its own registered projects.
    Commit ebe23b0 ("filter all notifications by registered projects") added
    a second-layer filter inside process_single_notification and
    _try_assignment_notification, but those branches still call
    mark_notification_read() for unknown repos — a write to shared GitHub
    state that clears the notification from the inbox of the sibling
    instance that actually owns the repo.

    The bug surfaces whenever the fetch_unread_notifications known_repos
    filter is bypassed, which happens when _get_known_repos_from_projects
    returns None (no project has github_url set, cold start before
    ensure_github_urls runs, or lazy workspace-cache population).

    Fix: add a single ownership gate at the top of _process_one_notification
    (loop_manager.py) that early-returns when
    resolve_project_from_notification yields None, and strip the redundant
    mark_notification_read calls from the two inner foreign-repo branches.
    The three defense layers (fetch filter, ownership gate, inner branches)
    now agree on the same invariant: never write to shared GitHub state for
    a repo this instance does not own.

### DIFF
--- a/koan/app/github_command_handler.py
+++ b/koan/app/github_command_handler.py
@@ -530,6 +530,40 @@ def resolve_project_from_notification(notification: dict) -> Optional[Tuple[str,
     return project_name, owner, repo
 
 
+def _skip_if_foreign_repo(
+    notification: dict, log_prefix: str,
+) -> Optional[Tuple[str, str, str]]:
+    """Resolve the project for ``notification`` or log a foreign-repo skip.
+
+    Centralizes the resolve-or-log boilerplate that previously lived in
+    ``process_single_notification``, ``_try_assignment_notification`` and
+    ``_try_subscription_notification``. Callers decide what to return on
+    a miss (``False``, ``(False, None)``, etc.) — this helper only does
+    the resolution and the debug log.
+
+    Args:
+        notification: A notification dict.
+        log_prefix: Short label included in the debug log so the source of
+            the skip is visible in ``/logs`` (e.g. ``"GitHub"`` for the
+            command path, ``"GitHub assign"`` for the assignment path).
+
+    Returns:
+        ``(project_name, owner, repo)`` when the repo is registered to
+        this instance, ``None`` otherwise.
+    """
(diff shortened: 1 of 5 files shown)


atoomic:
@Koan-Bot review

koan/app/github_command_handler.py
```diff
@@ -945,12 +945,14 @@ def _try_assignment_notification(
...
-    # Resolve project
+    # Resolve project — do NOT mark the notification as read when the repo
+    # isn't registered to this instance. Marking-read is a write to shared
+    # GitHub state; doing it here would hide the notification from a sibling
+    # Kōan instance that owns the repo.
     project_info = resolve_project_from_notification(notification)
     if not project_info:
         repo_name = notification.get("repository", {}).get("full_name", "?")
         log.debug("GitHub assign: repo %s not in projects.yaml", repo_name)
-        mark_notification_read(notif_id)
```
atoomic, line 953: @Koan-Bot ask is there a risk of accumulation of unwanted notifications that would prevent us to view the ones that matter to our project

koan/app/github_command_handler.py
```diff
@@ -1069,14 +1071,17 @@ def process_single_notification(
...
+    # own the repo. The outer _process_one_notification already gates the
+    # mark-read on the same ownership check.
     project_info = resolve_project_from_notification(notification)
     if not project_info:
         repo_data = notification.get("repository", {})
         full_name = repo_data.get("full_name", "?")
         reason = notification.get("reason", "?")
         log.debug("GitHub: repo %s (reason=%s) not in projects.yaml — ignoring notification", full_name, reason)
-        mark_notification_read(str(notification.get("id", "")))
         return False, None
```
atoomic, line 1110: looks like a boilerplate from the block above and room for improvement / refactoring

PR #1558, opened 2026-05-26
fix(review): fall back to local git diff when PR has > 300 files

atoomic:
GitHub's pr-diff endpoint returns HTTP 406 for diffs with more than 300 changed files, which made /review fail with the misleading "PR has no diff — nothing to review" message on huge PRs. The underlying gh error was also silently swallowed, so operators had no clue what failed.

Fix both:
- fetch_pr_context() now logs the gh error message and classifies it as "oversized" when it matches the 406/too_large signature.
- On 406, when project_path is available, fall back to a local `git fetch <url> pull/<N>/head` + `git diff base...head` from the checkout. Git has no file-count cap, so PRs of any size now produce a diff that the review pipeline (and compressor) can work with.
- review_runner threads project_path through to fetch_pr_context so the fallback path is reachable from /review.

### DIFF
--- a/koan/app/rebase_pr.py
+++ b/koan/app/rebase_pr.py
@@ -149,6 +149,25 @@ def _diff_too_large(error_message: str) -> bool:
     return any(marker in error_message for marker in _DIFF_TOO_LARGE_MARKERS)
 
 
+def _token_fetch_url(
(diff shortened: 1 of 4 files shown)


atoomic:
@Albertia-Bot review

PR #1563, opened 2026-05-26
fix(review): use review_mode for review skill + improve codex

atoomic:
* fix(review): capture Codex final output reliably
* fix(review): use review_mode for review skill


atoomic:
@Albertia-Bot review

PR #1567, opened 2026-05-26
fix: update codex full access flag

atoomic:
Replace deprecated --yolo with --dangerously-bypass-approvals-and-sandbox for Codex full-access mode. Warn at mission start when Codex runs without skip_permissions in implement/deep modes, since workspace-write may mount .git read-only and block branch/commit/push/PR operations.


atoomic:
@Albertia-Bot review

PR #1592, opened 2026-05-28
Notification polling guard

atoomic:
Improved experience when pause mode is disabled


atoomic:
@Koan-Bot review

atoomic:
checked and confirmed working as expected on a server with pause disabled

PR #1628, opened 2026-05-29
feat(jira): post structured PR/plan comments back to Jira issues

atoomic:


atoomic:
Issue -1- about the body is a nice optimization we could add along the road, not really a blocker 
confirmed it works fine

PR #1663, opened 2026-06-02
fix: load .env before git identity setup in startup_manager

exciton:
## Problem

`setup_git_identity()` reads `KOAN_EMAIL` from the environment to set `GIT_AUTHOR_EMAIL` and `GIT_COMMITTER_EMAIL`. However, `run.py` never sources the `.env` file before calling `run_startup()`, so `KOAN_EMAIL` is not present in the environment at the time git identity is configured. Git silently falls back to the global `~/.gitconfig` identity, causing commits to be authored under the system user rather than the configured Kōan identity.

## Fix

Add a `load_dotenv()` call in `run_startup()` after the pause/passive setup phase but before `setup_git_identity()`. `load_dotenv()` uses `os.environ.setdefault`, so it will not overwrite variables that were already set explicitly in the environment.

## Testing

- With `KOAN_EMAIL` set only in `.env` (not exported in the shell), git commits now use the configured identity.
- When `KOAN_EMAIL` is already in the environment, `setdefault` is a no-op and behavior is unchanged.

### DIFF
--- a/koan/app/startup_manager.py
+++ b/koan/app/startup_manager.py
@@ -518,6 +518,12 @@ def run_startup(koan_root: str, instance: str, projects: list):
     _safe_run("Start on pause", handle_start_on_pause, koan_root)
     _safe_run("Start passive", handle_start_passive, koan_root)
 
+    # Load .env before git identity so KOAN_EMAIL is available.
+    # run.py does not source .env itself; without this, GIT_AUTHOR_EMAIL is
+    # never set and git falls back to the global ~/.gitconfig identity.
+    from app.utils import load_dotenv
+    load_dotenv()
+
     # Git identity and GitHub auth
     _safe_run("Git identity", setup_git_identity)
     _safe_run("GitHub auth", setup_github_auth)


koan/app/startup_manager.py
```diff
@@ -518,6 +518,12 @@ def run_startup(koan_root: str, instance: str, projects: list):
     _safe_run("Start on pause", handle_start_on_pause, koan_root)
     _safe_run("Start passive", handle_start_passive, koan_root)
 
+    # Load .env before git identity so KOAN_EMAIL is available.
+    # run.py does not source .env itself; without this, GIT_AUTHOR_EMAIL is
+    # never set and git falls back to the global ~/.gitconfig identity.
+    from app.utils import load_dotenv
+    load_dotenv()
```
atoomic, line 525: Are we setting the env twice now? 
Could this result in leaking some unwanted env later to some runners?

exciton: Claude explanation: 

Here's the execution path for each relevant make command:

  make run → app/run.py

```
  run.py: main()
    └─ main_loop()
         └─ run_startup(koan_root, instance, projects)   # line 819
              └─ startup_manager.run_startup()           # line 501
                   └─ load_dotenv()                      # line 503–504
```

  run.py itself never calls load_dotenv() at module level or in main()/main_loop(). The one and only call is inside startup_manager.run_startup(), late in the startup sequence — deliberately placed just before setup_git_identity() so
  KOAN_EMAIL is available for git config.

  make awake → app/awake.py

```
  awake.py: imports bridge_state                         # line 29
    └─ bridge_state.py (module-level):
         └─ load_dotenv()                                # line 17
```

  awake.py triggers load_dotenv() via its top-level from app.bridge_state import …. bridge_state calls it at module level so that KOAN_TELEGRAM_TOKEN, KOAN_TELEGRAM_CHAT_ID, etc. are resolved before the constants are bound. awake.py's own
  main() never calls it again.

  make start → app.pid_manager start-all

  pid_manager.start_all() launches run.py and awake.py as separate subprocesses, so each path above applies independently in its own process — no double-load.

  ---
  Why there's no double-load

  The key is os.environ.setdefault() inside load_dotenv() (utils.py:104):

  os.environ.setdefault(key.strip(), value.strip().strip("\"'"))

  setdefault is a no-op if the key already exists. So even if load_dotenv() were called twice in the same process (e.g., from a messaging helper later), it would not overwrite already-set env vars. Each subprocess gets its own
  environment, so the run and awake processes are fully independent.

  Summary: run.py calls load_dotenv() exactly once (via startup_manager, near the end of startup). awake.py calls it exactly once (via bridge_state at import time). No process calls it twice in a way that matters, and setdefault makes
  double-calls safe anyway.

PR #1678, opened 2026-06-02
fix(usage): record skill-dispatch activity + backfill historical days

atoomic:
The dashboard /usage page reads daily mission activity solely from instance/usage/*.jsonl. Skill-dispatch missions (e.g. /review) produce runner text instead of CLI JSON, so token extraction returned None and _record_cost_event silently wrote no row — leaving recent days blank on the dashboard even though session_outcomes.json had the activity.

Forward fix (already in tree, now landed): capture token usage from provider stream events into a sidecar file (provider/__init__.py + run.py), and write a zero-token placeholder row for skill-dispatch runs when extraction still fails (mission_runner.allow_placeholder), so a row is recorded for every run. Also sort dashboard PRs by updatedAt.

Backfill: add `python -m app.backfill_usage` (dry-run by default, --apply to write) to reconstruct historical per-day activity from session_outcomes.json. Synthetic rows carry a "backfill": true marker (zero tokens/cost — counts only), and the tool is idempotent: it credits existing real and prior backfill rows so days are never over-counted.

### DIFF
--- a/docs/operations/dashboard.md
+++ b/docs/operations/dashboard.md
@@ -27,14 +27,32 @@ config) and exposes a JSON/SSE API under `/api/*`.
 | `/` | Dashboard — agent status, mission counts, attention zone, health, projects |
 | `/missions` | Pending / in-progress / done missions, with drag-reorder, edit, cancel |
 | `/chat` | Chat with the agent or queue a mission |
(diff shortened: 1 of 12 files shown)


atoomic:
@Koan-Bot review

atoomic:
@Koan-Bot rebase important

PR #1739, opened 2026-06-04
fix(awake): don't sys.exit(1) on non-telegram messaging providers

troglodyne:
## Summary

Makes the `awake` bridge work for non-Telegram messaging providers (Slack/Matrix, and any provider where `CHAT_ID` is unset) instead of crash-looping on the first message. The bridge's main loop and config check had baked in several Telegram-Bot-API assumptions that broke other providers.

## Changes

**Startup / config**
- `check_config()` no longer hardcodes a `KOAN_TELEGRAM_TOKEN`/`CHAT_ID` presence check that exited non-telegram setups. The telegram-only credential check is gated to `resolve_provider_name() == "telegram"`; each provider's own `configure()` does its real credential check.
- Guard the `BOT_TOKEN` / `CHAT_ID` startup log lines against empty values (no more slicing an empty token or logging an empty chat).
- Unhardcode the bridge banner provider name (was stuck on `"telegram"` regardless of the active provider).

**Main-loop crash fixes** (Matrix crash-looped on every message)
- Matrix HTTP fallback now mints a Telegram-shaped wrapper (`update_id`, `message.chat.id`, …) in `_parse_room_events` instead of returning raw Matrix events, which raised `KeyError: 'update_id'`.
- Bind `message_id` / mention-stripping inside the channel-matching block so it's set whenever `chat_id` matches the active channel (previously only inside a `CHAT_ID` guard → `UnboundLocalError` on Matrix).
- Guard the `offset = update["update_id"] + 1` access; non-telegram providers may omit `update_id`.

**Review follow-ups**
- Chat-id filter no longer treats an empty `CHAT_ID` as a wildcard: strip `""` from the match set and require a non-empty `chat_id`, so a malformed update missing `chat.id` can't slip past the channel filter.
- Promote `_resolve_provider_name` to a public `resolve_provider_name()` (private alias kept for backward compat) and use it in `check_config()` and `main()`.

## Tests

Added regression coverage in `koan/tests/test_awake.py`:
- non-telegram `check_config()` does not exit (telegram without creds still does)
- Matrix-shaped `main()` loop drives through without crashing
- chat-id filter positive/negative + malformed-update regression

## Files

- `koan/app/awake.py`
- `koan/app/messaging/__init__.py`
- `koan/app/messaging/matrix.py`
- `koan/tests/test_awake.py`

### DIFF
--- a/koan/app/awake.py
+++ b/koan/app/awake.py
@@ -83,7 +83,12 @@ def _get_last_message_id() -> int:
 
 
 def check_config():
-    if not BOT_TOKEN or not CHAT_ID:
+    # BOT_TOKEN / CHAT_ID are Telegram-specific.  Slack and Matrix users
+    # don't set them — defer the actual credential check to each
+    # provider's own ``configure()`` (called from get_messaging_provider
+    # below) so non-telegram providers don't get sys.exit(1)'d here.
+    from app.messaging import resolve_provider_name
+    if resolve_provider_name() == "telegram" and (not BOT_TOKEN or not CHAT_ID):
         log("error", "Set KOAN_TELEGRAM_TOKEN and KOAN_TELEGRAM_CHAT_ID env vars.")
         sys.exit(1)
     if not INSTANCE_DIR.exists():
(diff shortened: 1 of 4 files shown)


troglodyne:
Found this while using bdraco's matrix backend logic. Have a branch for enabling e2ee but it requires patching the megolm lib to work with latest element client. Picked this out of that branch for submission since this is a good fix regardless.

atoomic (approved):

troglodyne:
Added second commit that fixes an explosion when the matrix message *does not look like* a telegram message to the backend and frequently crashes.

troglodyne:
Hah. I think I actually hit that now after queueing a mission to fix what was pointed out in the review here. Fixing now.

atoomic (approved):

PR #1799, opened 2026-06-06
feat(chat): intent routing with confirmation for natural-language actions

### LINKED ISSUE #1798: Permission blocker: autonomous agents should not request approval for instance/* runtime files
## Problem

When Kōan (or any autonomous agent) attempts to modify runtime queue files like `instance/missions.md`, it must request explicit user approval via the Claude Code permission system. This defeats the autonomy model.

`instance/missions.md` is not source code — it's the agent's runtime task queue (Pending/In Progress/Done sections). The agent **must** be able to update it atomically without blocking on permission dialogs.

## Root cause

The Claude Code harness treats all file modifications equally. Kōan's autonomy requires a distinction:
- **Source code** (under `koan/`, `docs/`, etc.): require approval
- **Runtime state** (under `instance/`): should not require approval

## Solution

Configure permission allowlist in `.claude/settings.json` (or equivalent) to auto-allow writes to `instance/*` files. This includes:
- `instance/missions.md` — task queue
- `instance/outbox.md` — message buffer  
- `instance/journal/*` — daily logs
- `instance/memory/*` — learnings and context
- `instance/.*.json` — state files

Alternatively, expose a Telegram command that configures permissions persistently so the operator can enable it once.

## Impact

- Blocks recurring task restarts, mission queuing, and state persistence
- Forces operator to be present for every autonomous decision (anti-pattern for a background agent)

sukria-koan0:
## Summary

Implements natural-language intent routing for chat. Users can now send conversational requests like 'restart my recurring tasks' and Kōan will propose the matching slash command (`/resume_recurring`) with a confirmation prompt.

Closes https://github.com/Anantys-oss/koan/issues/1798

## Changes

- **Phase 1**: Confirmation state management (pending_action store)
- **Phase 2**: Intent classifier prompt and _classify_intent helper
- **Phase 3**: Chat intent routing with confirm/reject flow
- **Phase 4**: Recurring missions as canonical example
- **Phase 5**: Operations documentation

## Tests

22 new unit tests covering all phases; all pass.

---

### DIFF
--- a/docs/operations/permissions.md
+++ b/docs/operations/permissions.md
@@ -1,0 +1,89 @@
+# Claude Code Permissions & Kōan Runtime Files
+
+## Overview
+
+Kōan manages its own runtime state (`instance/missions.md`, `instance/recurring.json`, `instance/soul.md`, etc.) via Python atomic writes in the bridge (`awake.py`) and agent loop (`run.py`). **These operations do NOT go through Claude Code's permission system** — they are safe-by-design and require no allowlist configuration.
+
+When the chat system detects an actionable intent (e.g., "restart my recurring tasks"), it routes the command through existing Python handlers, not through Claude's file-write tools. This keeps the chat tool list read-only (Read, Glob, Grep) to prevent Telegram prompt-injection attacks, while still enabling state-changing operations via confirmed commands.
+
+## When Permissions Don't Apply
+
+- **Bridge and agent loop**: Kōan's own Python code reads and writes `instance/*` files directly. No Claude Code calls involved.
+- **Chat system**: Intent routing and confirmation happen in Python; dispatch goes through validated handlers.
+- **Mission execution**: Missions run via the agent loop, not chat; only the agent loop needs Claude access, and it already has unrestricted permission to your projects' source code.
+
+## When Permissions Do Apply
+
+**Only when a mission is authored to edit the Kōan codebase itself** (e.g., "add a new core skill", "fix a bug in awake.py").
+
+If a mission targets a workspace project other than Kōan, or if it reads Kōan code, permissions are not involved. Permissions only matter when a mission task would run Claude Code with `Edit` or `Write` against Kōan's `instance/` or `koan/` source directories.
+
+## Optional: Allowlist Configuration
+
+### For missions that edit the Kōan repo itself
+
+If you want to author missions that improve Kōan's own code (fixing bugs in the bridge, adding skills, etc.), you have two options:
+
+#### Option 1: Skip permissions (simple, per-mission)
+
+In your mission text, add the `[skip_permissions]` tag:
+
+```
+mission: [skip_permissions] add a new core skill for X
+```
+
+This tells the agent loop to run Claude without the permission dialog. Useful for one-off tasks where you trust the content.
+
+**Default:** `skip_permissions: false` (enabled by default is not recommended for security).
+
+#### Option 2: Allowlist `instance/` files (recommended, permanent)
+
+Add a rule to `.claude/settings.json` in your Kōan directory to allow Claude to write `instance/` files:
+
+```json
+{
+  "permissions": {
+    "write": [
+      "instance/**"
+    ]
+  }
+}
+```
+
+This allows Claude to edit all runtime files in `instance/` (missions, recurring, memory, etc.) without a permission prompt.
+
+**Note:** This only applies to the Kōan repo itself. Individual workspace projects have their own `.claude/settings.json` and are unaffected.
+
+### Operator-specific: Don't commit allowlist to the public repo
+
+If you maintain a custom Kōan fork and want to add the allowlist, use `.claude/settings.local.json` (git-ignored) instead of `.claude/settings.json` so the allowlist doesn't leak to the public repository:
+
+```bash
+cat >> .claude/settings.local.json <<'EOF'
+{
+  "permissions": {
+    "write": ["instance/**"]
+  }
+}
+EOF
+```
+
+## FAQ
+
+**Q: Why does chat not have Write access?**  
+A: Chat runs via Telegram, which is an untrusted channel. An attacker could send a prompt-injection payload and trick Claude into editing your mission list. By keeping chat read-only and routing all actions through Python validation, we prevent this attack. Confirmed commands still execute via Python handlers, bypassing the need for Write tools.
+
+**Q: Do I need a permission allowlist to run the agent loop?**  
+A: No. The agent loop is Python code in `koan/app/run.py`, not a Claude Code call. It uses atomic writes to manage `instance/` files directly.
+
+**Q: Why does `instance/missions.md` never prompt for permissions?**  
+A: Because Kōan's Python code writes it directly via `app.utils.atomic_write()` in `missions.py`, `start_mission()`, etc. This happens in the bridge/agent, not in a Claude Code session.
+
+**Q: Can I write my own missions against the Kōan repo?**  
+A: Yes, if you configure the allowlist or use `[skip_permissions]`. This is the intended flow for operators who want Kōan to improve itself — add a mission like "implement [feature]" with the tag, and Kōan will author the code.
+
+## See Also
+
+- [Prompt Guard & Injection Mitigation](../security/prompt-guard.md)
+- [Agent Loop & Mission Execution](../architecture/daemon.md)
+- [Bridge Architecture](../architecture/overview.md)
--- a/docs/operations/permissions.md
+++ b/docs/operations/permissions.md
@@ -24,17 +24,17 @@ If a mission targets a workspace project other than Kōan, or if it reads Kōan 
 
 If you want to author missions that improve Kōan's own code (fixing bugs in the bridge, adding skills, etc.), you have two options:
 
-#### Option 1: Skip permissions (simple, per-mission)
+#### Option 1: Skip permissions globally (simple)
 
-In your mission text, add the `[skip_permissions]` tag:
+Set `skip_permissions: true` in `instance/config.yaml`:
 
+```yaml
+skip_permissions: true
 ```
-mission: [skip_permissions] add a new core skill for X
-```
 
-This tells the agent loop to run Claude without the permission dialog. Useful for one-off tasks where you trust the content.
+This adds `--dangerously-skip-permissions` to the Claude CLI invocation (see `get_skip_permissions()` in `config.py`), so the agent loop runs Claude without the permission dialog. Useful when you trust all mission content.
 
-**Default:** `skip_permissions: false` (enabled by default is not recommended for security).
+**Default:** `skip_permissions: false` (disabling permissions is not recommended for security; prefer the allowlist in Option 2).
 
 #### Option 2: Allowlist `instance/` files (recommended, permanent)
 
--- a/koan/app/awake.py
+++ b/koan/app/awake.py
@@ -375,17 +375,39 @@ def handle_chat(text: str, chat_id: Optional[str] = None):
     # Save user message to history
     save_conversation_message(CONVERSATION_HISTORY_FILE, "user", text)
 
+    # Scan for prompt injection — warn-only (never block chat; tools are read-only).
+    # Runs BEFORE the confirmation/intent shortcuts so an injected "yes" or a
+    # crafted actionable phrase cannot skip the guard entirely (review warning #6).
+    from app.prompt_guard import scan_mission_text
+    from app.config import get_prompt_guard_config
+    from app.command_handlers import quarantine_mission
+
+    guard_config = get_prompt_guard_config()
+    if guard_config["enabled"]:
+        guard_result = scan_mission_text(text)
+        if guard_result.blocked:
+            log("guard", f"WARNING chat: {guard_result.reason} | {text[:100]}")
+            quarantine_mission(text, guard_result.reason, source="telegram-chat")
+
     # --- Confirmation flow: check if user is confirming or rejecting a pending action ---
     if _is_confirmation(text) or _is_rejection(text):
         pending = get_pending_action(chat_id)
         if pending:
             if _is_confirmation(text):
-                # User confirmed — dispatch the command
-                clear_pending_action(chat_id)
-                log("intent", f"Confirmed: {pending['command']}")
-                # Dispatch via command_handlers (the safe path that validates everything)
+                # User confirmed — dispatch the command. Keep the pending action
+                # until dispatch succeeds so a failure can be retried (warning #5).
+                command = pending["command"]
+                log("intent", f"Confirmed: {command}")
                 from app.command_handlers import handle_command
-                handle_command(pending["command"])
+                try:
+                    handle_command(command)
+                except Exception as e:
+                    log("error", f"Confirmed command failed: {command} | {e}")
+                    err = f"⚠️ Couldn't run `{command}`: {e}"
+                    send_telegram(err)
+                    save_conversation_message(CONVERSATION_HISTORY_FILE, "assistant", err)
+                    return
+                clear_pending_action(chat_id)
                 return
             else:  # Rejection
                 # User rejected — clear and acknowledge
@@ -393,27 +415,47 @@ def handle_chat(text: str, chat_id: Optional[str] = None):
                 send_telegram("❌ Cancelled.")
                 save_conversation_message(CONVERSATION_HISTORY_FILE, "assistant", "❌ Cancelled.")
                 return
+        elif _is_confirmation(text):
+            # Confirmation word with nothing pending (e.g. expired) — give feedback
+            # instead of silently routing "yes" through normal chat (Phase 1).
+            msg = "Nothing pending to confirm."
+            send_telegram(msg)
+            save_conversation_message(CONVERSATION_HISTORY_FILE, "assistant", msg)
+            return
 
     # --- Intent classification: check if this looks like an actionable request ---
     # Cheap pre-filter first: only pay the ~5-10s classifier cold-start when the
     # message actually mentions a known command keyword. Plain chat skips it.
     intent = _classify_intent(text) if _mentions_known_command(text) else None
     if intent and intent.get("actionable") and intent.get("command"):
-        confidence = intent.get("confidence", 0)
-        # High-confidence, low-risk commands: execute without confirmation
-        # Risk assessment: /status, /list, /show are safe; anything else needs confirmation
-        safe_commands = {"/status", "/list", "/show", "/help", "/projects"}
-        command_name = intent["command"].split()[0]  # Extract /command part
-        is_high_confidence = confidence >= 0.7
-        is_safe = any(intent["command"].startswith(sc) for sc in safe_commands)
+        # Absent confidence is treated as non-actionable rather than silently
+        # defaulting to 0 (which would fail every threshold check) (Phase 4).
+        if "confidence" not in intent:
+            log("intent", f"Classifier output missing 'confidence'; ignoring: {intent.get('command')}")
+            intent = None
+        else:
+            confidence = intent["confidence"]
+            # High-confidence, low-risk commands: execute without confirmation.
+            # Risk assessment: /status, /list, /show are safe; else needs confirmation.
+            safe_commands = {"/status", "/list", "/show", "/help", "/projects"}
+            command_name = intent["command"].split()[0]  # Extract /command part
+            is_high_confidence = confidence >= 0.7
+            is_safe = command_name in safe_commands
 
-        if is_high_confidence and is_safe:
-            # Auto-execute safe, high-confidence commands
-            log("intent", f"Auto-executing (high-confidence safe): {intent['command']}")
-            from app.command_handlers import handle_command
-            handle_command(intent["command"])
-            return
+            if is_high_confidence and is_safe:
+                # Auto-execute safe, high-confidence commands
+                log("intent", f"Auto-executing (high-confidence safe): {intent['command']}")
+                from app.command_handlers import handle_command
+                try:
+                    handle_command(intent["command"])
+                except Exception as e:
+                    log("error", f"Auto-execute failed: {intent['command']} | {e}")
+                    err = f"⚠️ Couldn't run `{intent['command']}`: {e}"
+                    send_telegram(err)
+                    save_conversation_message(CONVERSATION_HISTORY_FILE, "assistant", err)
+                return
 
+    if intent and intent.get("actionable") and intent.get("command") and "confidence" in intent:
         # All other actionable commands: propose with confirmation
         store_action = {
             "command": intent["command"],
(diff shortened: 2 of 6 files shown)


sukria:
Looks very good.
To be tested.

sukria:
@sukria-koan0 plan 

(you want to address all Silent Failure Analysis issues reported above.

atoomic:
@Koan-Bot ask I wonder if the issue here is a regression from the recent implementation 19a0aa8cb6a0f66ec884e7a889b11d9120b3cdf6

sukria:
@atoomic it is not. It is a new feature which is really nice to have. And it works fine.
Let's discuss this when we can.

sukria:
@sukria-koan0 rebase to address the security concern spotted

sukria-koan0:
## Address all four Silent Failure Analysis issues in chat intent routing (PR #1799)

### Changes in this iteration

- **Scoped the iteration to the four Silent Failure Analysis findings** flagged by the automated review on PR #1799 (branch `7cf85524`), per sukria's instruction "address all Silent Failure Analysis issues reported above." The broader PR-review warnings (vocabulary pre-filter, guard ordering, etc.) are tracked separately by the `rebase` request and are noted but not the focus here.
- **Verified the live code** on the PR branch rather than `main` (the intent-routing code is not yet merged). Confirmed exact current shape of `handle_chat` (`koan/app/awake.py:359`), `_classify_intent` (`:589`), and the confirmation block (`:378-415`). Line numbers in the review are slightly off the branch HEAD; the plan references current branch lines.
- **Phase 1 (expired-confirmation no-op)**: confirmed the `if pending:` block at `:380` has no `else`, so a `yes`/`no` after the 120s TTL falls through to conversational chat. Plan adds an explicit "nothing pending" reply.
- **Phase 2 (clear-before-execute)**: confirmed `clear_pending_action` runs at `:384`/`:391` before `handle_command`. Plan wraps dispatch in try/except and preserves the action on failure.
- **Phase 3 (catch-all)**: confirmed `except (json_lib.JSONDecodeError, subprocess.TimeoutExpired, Exception)` at `:659` — `Exception` makes the tuple a bare catch-all. Plan narrows it to `OSError` + the two specific types.
- **Phase 4 (missing-confidence)**: confirmed `confidence = intent.get("confidence", 0)` at `:402` silently defaults to 0; and `_classify_intent` validation at `:657` already requires `"confidence" in intent`, so the real gap is a silent return with no log. Plan adds logging and tightens validation.

### Summary

PR #1799 introduces natural-language intent routing in `handle_chat`. The automated Silent Failure Analysis found four paths where failures or malformed inputs degrade silently: an expired confirmation produces no feedback, pending state is consumed before the fallible command runs, a catch-all swallows real bugs in the classifier, and unvalidated LLM output silently fails the confidence gate. This plan makes each of those paths observable and recoverable without changing the happy-path behavior, so the feature is safe to test in production.

### Alternatives Considered

- **Approach A (chosen)**: Targeted fixes to the four flagged sites in `awake.py`, keeping the existing control flow. *Trade-off: minimal blast radius and easy review, but doesn't restructure the broader guard-ordering / auto-execute concerns (left to the separate rebase).*
- **Approach B**: Restructure `handle_chat` into a small dispatch pipeline (guard → confirmation → intent → chat) that centralizes error handling and conversation-history writes. *Trade-off: cleaner long-term and would also fix the guard-bypass warning, but larger diff and re-touches the 22 existing tests — out of scope for "address Silent Failure issues."*
- **Approach C**: Move pending-action storage and TTL handling into `bridge_state.py` with a single `consume_pending_action()` that returns expired/valid status. *Trade-off: better separation, but premature until the confirmation flow stabilizes.*

### Implementation Phases

#### Phase 1: Feedback on expired/absent confirmation

- **What**: In `handle_chat` (`koan/app/awake.py`), add an `else` branch to the `if pending:` block (currently ends at `:399`). When the user sends `yes`/`no` but `get_pending_action` returns `None`, send "Nothing pending to confirm (it may have expired)." and `save_conversation_message`, then `return`.
- **Why**: Without it, `yes` falls through and Claude answers the literal word conversationally, appearing to ignore the confirmation.
- **Gotchas**: Only short-circuit when the message was actually classified as confirmation/rejection (i.e., inside the existing `if _is_confirmation(text) or _is_rejection(text):` guard) — don't intercept normal chat that happens to contain "no".
- **Done when**: Sending `yes` with no pending action returns the "nothing pending" message and does not invoke the chat CLI.

#### Phase 2: Preserve pending action across dispatch failure

- **What**: In the confirmation branch (`:382-392`), move `clear_pending_action(chat_id)` to after a successful `handle_command(pending["command"])`, wrapped in try/except. On exception: keep the pending action, send an error message, save it to conversation history, and log. Also add `save_conversation_message` on the success path (success is currently invisible to chat history).
- **Why**: Clearing before dispatch makes a raised command irrecoverable — no retry, no feedback.
- **Gotchas**: `handle_command` is imported lazily inside the branch; keep that. Decide TTL behavior on retry — keep the original `expires_at` so a hung confirmation still expires (don't refresh it).
- **Done when**: When `handle_command` raises, the user receives an error message, the pending action is still retrievable via `get_pending_action`, and a successful confirmation writes an assistant entry to conversation history.

#### Phase 3: Narrow the classifier exception handler

- **What**: In `_classify_intent` (`:659`), change `except (json_lib.JSONDecodeError, subprocess.TimeoutExpired, Exception) as e:` to `except (json_lib.JSONDecodeError, subprocess.TimeoutExpired, OSError) as e:`. Let unexpected exceptions (NameError/TypeError/AttributeError/KeyError) propagate.
- **Why**: `Exception` is a superclass of the two specific types, so the current tuple is a bare catch-all that degrades programming bugs to a "no intent" log line — a broken classifier would look like it's working.
- **Gotchas**: Confirm no caller depends on `_classify_intent` never raising. `handle_chat` calls it at `:400`; since propagation here means a genuine bug, surfacing it in the run loop's logs is the intent — but verify it won't crash the bridge poll loop. If the bridge loop wraps `handle_chat` in a catch-all (it does at `awake.py` outer loop), propagation is safe and surfaces in logs. Note `test_silent_exceptions` enforces all catch blocks log — this block already logs, so it stays compliant.
- **Done when**: A simulated `TypeError` inside the JSON-parse block propagates out of `_classify_intent` (not swallowed), while a malformed-JSON / timeout / OSError still returns `None` with a log line.

#### Phase 4: Surface unvalidated / low-quality LLM output

- **What**: Two changes. (a) In `_classify_intent` validation (`:657`), when JSON parses to a dict but is missing `actionable` or `confidence`, log a warning ("classifier output missing required fields: …") before returning `None` instead of silently dropping it. (b) In `handle_chat` (`:402`), if `intent.get("confidence")` is absent, log a warning rather than defaulting to `0` silently; treat absence as non-actionable.
- **Why**: A model returning a valid actionable command without a `confidence` field would never propose or execute, with no diagnostic.
- **Gotchas**: Keep the `>= 0.7` threshold semantics. The validation in `_classify_intent` already requires `"confidence" in intent`, so item (b) is defense-in-depth — keep both but ensure they log, not just drop.
- **Done when**: A classifier response missing `confidence` produces a warning log and is treated as non-actionable; the message falls through to normal chat.

### Corner Cases

- User sends `yes` twice rapidly: first consumes/clears, second hits the new Phase 1 "nothing pending" path.
- Confirmation arrives exactly at TTL boundary — rely on `get_pending_action` returning `None` for expired entries (verify it enforces `expires_at`); Phase 1 then handles feedback.
- `handle_command` raises *and* the pending action TTL has since passed — Phase 2 re-stores it but it's already expired, so the next `yes` gets Phase 1 feedback (acceptable).
- Classifier returns valid JSON that is a list, not a dict — existing `isinstance(intent, dict)` guard handles; ensure no log spam.
- Empty/whitespace classifier stdout (`"{" not in text_output`) — still returns `None` silently; acceptable (no LLM call failure), but optionally log at debug.

### Testing Strategy

- Run with `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_awake.py -v` (plus `test_bridge_state.py` for pending-action TTL).
- Mock `_classify_intent`, `handle_command`, and `send_telegram`; never invoke the real CLI. Mock `format_and_send` per CLAUDE.md test rules.
- New/updated tests:
  - `yes` with no pending action → "nothing pending" message sent, no chat CLI call (Phase 1).
  - `handle_command` raises on confirm → error message sent, `get_pending_action` still returns the action, success-path writes history (Phase 2).
  - Inject a `TypeError` into the JSON block → assert it propagates from `_classify_intent`; malformed JSON / `TimeoutExpired` / `OSError` → returns `None` and logs (Phase 3).
  - Classifier dict missing `confidence` → warning logged, falls through to normal chat (Phase 4).
- Add the autouse `_clear_pending` fixture (review warning #6) so module-level `_pending_actions` is reset between tests, preventing state leakage now that more tests touch it.
- `make lint` must pass (no bare-except reintroduced).

### Risks & Alternatives

- **Risk**: Narrowing the exception handler (Phase 3) could let a previously-swallowed bug crash the chat path. *Mitigation*: the bridge outer loop already catches `Exception` around message handling, so it logs and continues rather than killing the process; this is the desired observability.
- **Risk**: Preserving pending action on failure (Phase 2) could leave a stale action that the user keeps confirming into the same failure. *Mitigation*: original `expires_at` is retained, so it self-clears within 120s.
- **Out of scope (separate rebase request)**: guard-runs-after-confirmation bypass (review warning #5), vocabulary pre-filter over-matching (blocking #1), and the auto-execute path itself. These overlap with Phase 2/3 but are addressed by the `@sukria-koan0 rebase` thread, not this plan.

### Verification Criteria

- Given a pending action older than 120s, when the user sends `yes`, then a "nothing pending" message is sent and no chat CLI invocation occurs.
- Given a pending action and a `handle_command` that raises, when the user confirms, then an error message is sent and `get_pending_action(chat_id)` still returns the action.
- Given a successful confirmation, when `handle_command` returns, then `clear_pending_action` has run and an assistant message is in conversation history.
- Given a `TypeError` raised inside `_classify_intent`'s parse block, when it executes, then the exception propagates (is not returned as `None`).
- Given classifier JSON missing `confidence`, when `handle_chat` processes it, then a warning is logged and the message is handled as normal chat.
- Running `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_awake.py koan/tests/test_bridge_state.py -v` and `make lint` produces no failures.

### Open Questions

- Should the Phase 2 retry path also re-send the original confirmation prompt ("I can run this for you… yes/no"), or just an error message? Defaulting to an error message plus preserved action unless you prefer re-prompting.
- Confirm `get_pending_action` enforces `expires_at` (returns `None` when expired). If it does not, Phase 1 needs to add the TTL check at read time — flag for verification during implementation.

---

sukria-koan0:

Branch `koan0/implement-1798` was rebased onto `main` and review feedback was applied.

### Changes applied

- ed `command_name`.
- **Replaced bare `except Exception`** in `_classify_intent` (warning #5, blocking) with `(JSONDecodeError, TimeoutExpired, OSError)` so programming bugs propagate instead of being swallowed.
- **Confidence handling** (Silent Failure / Phase 4): `_classify_intent` logs a warning and drops classifier output missing required fields; `handle_chat` treats absent `confidence` as non-actionable instead of silently defaulting to `0`.
- **Feedback for confirmation with no pending action** (Phase 1): a `yes` with nothing pending now replies "Nothing pending to confirm." instead of silently routing to normal chat. Updated the corresponding test.
- **Fixed `permissions.md`** (warning #8, blocking): documented the real global `skip_permissions: true` config key instead of the nonexistent `[skip_permissions]` mission tag.
- **Test isolation** (warning #7, blocking): added autouse `_clear_pending` fixtures resetting `_pending_actions` in `test_bridge_state.py` and `test_awake.py`.

### Stats

```
6 files changed, 809 insertions(+), 11 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=PR #1801 only added resume/run subcommands to /recurring, but no commit on main implements the natur)
- Rebased `koan0/implement-1798` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/implement-1798` to origin

</details>

### CI status

---
_Automated by Kōan_

atoomic:
@Koan-Bot review

sukria:
What about changing the strategy here : remove entirely the gating and routing skill, and just add a section in the default chat prompt to suggest a matching skill if the user's input is not understood.

This would not cost another round trip to Claude and should integrate much more nicely.

Reset to main and try this approach.

@sukria-koan0 plan

sukria-koan0:
## Reset to main and replace chat intent-routing with prompt-level skill suggestions

### Changes in this iteration

- **Strategy pivot per sukria's latest comment** (2026-06-06): abandon the entire intent-routing/confirmation machinery from PR #1799 (classifier round-trip, `pending_action` store, vocabulary pre-filter, confirm/reject flow). sukria's reasoning: it adds a second Claude call (5-10s latency), and the review thread confirmed the pre-filter over-matches common English words (Koan-Bot Important #1), firing the classifier on most ordinary chat. atoomic's regression question (`19a0aa8c`) was resolved by sukria as "not a regression, new feature."
- **New direction**: keep `handle_chat` as the existing single-call path on `main`, and enrich the chat prompt (`koan/system-prompts/chat.md`) with a compact catalog of available slash commands plus an instruction: when a user's natural-language message maps to an existing command, *suggest* the exact `/command` in the reply rather than answering generically. No execution, no extra round-trip.
- **All prior Silent Failure / review warnings become moot** — they targeted code that will no longer exist after reset to `main`. The chat path stays read-only (`Read/Glob/Grep`), so the guard-ordering and auto-execute security concerns disappear entirely.
- Verified current `main` state: `handle_chat` (`koan/app/awake.py:357`) has no intent routing; `_build_chat_prompt` (`:173`) assembles the prompt via `load_prompt("chat", ...)` with named placeholders; `chat.md` has no skill catalog section today. The skill registry (`skills.py`) exposes `list_by_audience()` / `list_all()` with `.commands`, `.description`, `.group` per skill.

### Summary

This plan discards the intent-routing feature in favor of a zero-cost, prompt-only approach: Kōan's chat already runs one Claude call with read-only tools, so we simply teach that call to recognize when a conversational request corresponds to a real slash command and to surface that command to the human (e.g. "Sounds like you want `/resume_recurring` — want me to queue it?"). It removes a fragile classifier, an entire confirmation state machine, a security bypass surface, and 5-10s of added latency, while delivering the same user value — discoverability of the right command from natural language — with the human still firmly in control of execution.

### Alternatives Considered

- **Approach A (chosen)**: Reset to `main`, add a "Suggesting commands" section + compact command catalog to `chat.md`, injected via a new `{SKILLS_CATALOG}` placeholder built in `_build_chat_prompt`. *Trade-off: relies on the chat model to map intent in-context (no deterministic routing), but costs zero extra calls, integrates natively, and removes all the security/latency concerns the review raised.*
- **Approach B (rejected — the PR #1799 approach)**: Heuristic pre-filter → dedicated classifier Claude call → confirmation state machine → dispatch. *Trade-off: deterministic and auditable, but adds a round-trip on most messages, an ephemeral state store, and a guard-bypass surface; sukria explicitly asked to drop it.*
- **Approach C (rejected)**: A cheap local keyword/fuzzy matcher (no LLM) that appends a "did you mean /x?" suffix to chat replies. *Trade-off: deterministic and free, but resurrects the over-matching problem from Important #1 ("status", "check", "fix" are common words) and duplicates intent reasoning the chat model already does well in-context.*

### Implementation Phases

#### Phase 1: Reset branch to main, drop PR #1799 code

- **What**: Start from a clean branch off `main` (`koan/<n>-chat-command-suggestions`). Confirm none of the intent-routing artifacts exist: no `_classify_intent`, `_mentions_known_command`, `_is_confirmation`/`_is_rejection`, `pending_action` helpers in `bridge_state.py`, no intent classifier prompt file, and no confirmation tests in `test_awake.py`. If any landed on `main`, remove them; otherwise this is a no-op verification.
- **Why**: The new approach supersedes the entire feature; we build on the simple single-call `handle_chat`.
- **Gotchas**: PR #1799's branch was `koan0/implement-1798` — do **not** branch from it. Verify `main` HEAD does not contain merged intent code (the plan author noted it was unmerged). Check `bridge_state.py` for stray `_pending_actions` / `set_pending_action` / `get_pending_action` / `clear_pending_action` and remove if present, plus their autouse `_clear_pending` test fixtures.
- **Done when**: `grep -rn "_classify_intent\|_pending_actions\|_mentions_known_command" koan/` returns nothing, and `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_awake.py koan/tests/test_bridge_state.py -v` passes on the clean base.

#### Phase 2: Build a compact command catalog for the chat prompt

- **What**: Add a helper (e.g. `_build_command_catalog()` in `koan/app/awake.py`, or better, a reusable `command_catalog()` in `skills.py`) that returns a compact, plain-text list of human-triggerable commands as `\/command — description` lines, grouped by the existing help `group`. Source it from the skill registry, filtering to skills a human invokes from Telegram: `list_by_audience("bridge", "command", "hybrid")` (exclude `agent`-only), and include the hardcoded core commands from `command_handlers._CORE_COMMAND_HELP`. Keep one line per primary command (skip aliases) and truncate descriptions to ~80 chars.
- **Why**: The chat model needs to know which commands exist to suggest the right one. Reusing the same data `/help` renders keeps the catalog in sync automatically as skills are added/removed (avoids the "hardcoded list" anti-pattern in CLAUDE.md).
- **Gotchas**: Prompt-budget — there are ~80 core skills; a full catalog could be heavy. Mitigate by emitting only `name — short description` (one line each), grouped, and by **omitting the catalog in `lite=True`** retry mode (the timeout-fallback path in `_build_chat_prompt`). Do not leak private `instance/skills/<scope>` identifiers — they're fine to surface to the operator in their own chat, but keep the catalog driven off the registry, not hardcoded names. Build the catalog once per call; it's cheap (registry is already loaded).
- **Done when**: The helper returns a deterministic, grouped, newline-joined string for the loaded registry, and a unit test asserts a known core command (e.g. `/recurring` or `/status`) with its description appears, and that an `agent`-only skill does not.

#### Phase 3: Add the "Suggesting commands" section to chat.md

- **What**: Add a new `{SKILLS_CATALOG}` placeholder to `koan/system-prompts/chat.md` (after the filesystem-layout block, before the closing tone rules), wrapped in a clear instruction block. Instruction text (paraphrased): *"You have these slash commands available. When the human's message expresses an intent that maps to one of them, name the exact command in your reply and offer to run it — e.g. 'That maps to `/resume_recurring` — want me to queue it?'. Suggest at most one command, only when the mapping is clear; otherwise just answer conversationally. Never claim you executed it — the human runs commands themselves."* Then append the catalog. Update `_build_chat_prompt` to pass `SKILLS_CATALOG=...` (empty string when `lite=True`).
- **Why**: This is the core of the new strategy — surfacing command discoverability inside the existing chat call.
- **Gotchas**: `load_prompt` does plain `str.replace` substitution (confirmed safe — no `str.format`), so brace-containing descriptions won't break it. Keep the instruction tight so the model doesn't spam command suggestions on every message. Recurring missions is the canonical example sukria cares about (`/resume_recurring`) — make it the worked example in the prompt. Ensure the new placeholder is always supplied (even empty) or `load_prompt` leaves a literal `{SKILLS_CATALOG}` in output.
- **Done when**: Rendering the chat prompt for a non-lite message includes the catalog and instruction; the lite/retry render omits the catalog; a message like "restart my recurring tasks" plausibly elicits a `/resume_recurring` suggestion (validated via a prompt-content assertion, not a live CLI call).

#### Phase 4: Documentation

- **What**: Update `docs/users/user-manual.md` and `docs/users/skills.md` to describe that chat now suggests matching slash commands from natural language (replacing any mention of intent-routing/auto-confirmation added for #1799, if present). Update `docs/operations/` if a #1799 ops doc was added — remove or rewrite it. Note in the chat/messaging docs that suggestions are advisory; the human always runs the command.
- **Why**: CLAUDE.md mandates docs stay in sync with user-facing behavior changes.
- **Gotchas**: If PR #1799 added a `docs/operations/*intent*` file or `permissions.md` edits, reconcile them — the `skip_permissions` doc fix from the #1799 rebase was a real correction; preserve that part if it's already on `main`, but drop intent-routing references.
- **Done when**: Docs describe the prompt-suggestion behavior, contain no references to a classifier/confirmation flow, and `grep -rn "intent rout\|pending_action\|confirmation flow" docs/` returns nothing stale.

### Corner Cases

- User message clearly maps to a destructive-ish command (e.g. `/reset`, `/shutdown`) — the model should still only *suggest*, never execute (chat tools are read-only, so it physically can't; the prompt reinforces it).
- Ambiguous message matching multiple commands — instruct "suggest at most one, only when clear" to avoid suggestion spam.
- Message in French — suggestions must respect `get_language_instruction()`; the command token stays literal (`/resume_recurring`) but the surrounding offer is translated.
- `lite=True` timeout-retry path — catalog omitted to stay under budget; suggestions simply won't appear on the degraded retry (acceptable).
- Empty registry / no human-facing skills — catalog helper returns `""`; the prompt section degrades gracefully (instruction still harmless).
- Private custom skills under `instance/skills/<scope>` — appear in the operator's own chat catalog (expected), but the helper must source from the registry, never a hardcoded name list.

### Testing Strategy

- `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_awake.py -v` — never invoke the real CLI; assert on the **prompt string** produced by `_build_chat_prompt`, not on Claude output.
- New tests:
  - Catalog helper includes a known core command + description and excludes `agent`-only skills (Phase 2).
  - `_build_chat_prompt(text)` (non-lite) contains the catalog and the "suggest the matching command" instruction; `_build_chat_prompt(text, lite=True)` does **not** contain the catalog (Phase 3).
  - `{SKILLS_CATALOG}` placeholder is fully substituted (no literal brace token leaks into the rendered prompt).
- Per CLAUDE.md test rules: mock `format_and_send` / `run_cli`; patch `app.notify.format_and_send` too under `runpy` if used.
- `make lint` and `make test` must pass.

### Risks & Alternatives

- **Risk**: The chat model may over-suggest commands, nagging the user. *Mitigation*: prompt says "at most one, only when the mapping is clear, otherwise just answer"; tune wording if testing shows noise.
- **Risk**: Prompt bloat from the catalog increases token cost / latency on the single chat call. *Mitigation*: one compact line per command, grouped; omitted in lite retry. Far cheaper than #1799's extra round-trip.
- **Risk**: Model suggests a command that doesn't exist (hallucination). *Mitigation*: catalog is the authoritative in-context list; instruction says "from the list above." Low impact — a wrong suggestion is just text the human can ignore.
- **Out of scope**: deterministic auto-execution of suggested commands (deliberately dropped — violates "the human decides" and reintroduces the security surface).

### Verification Criteria

- Given a clean branch off `main`, when `grep -rn "_classify_intent\|_pending_actions\|_mentions_known_command" koan/` runs, then it returns no matches.
- Given the chat registry, when `_build_command_catalog()` runs, then its output contains `/recurring` (or `/status`) with a description and contains no `agent`-only skill.
- Given a non-lite chat prompt, when `_build_chat_prompt("restart my recurring tasks")` renders, then the output contains the command catalog and the command-suggestion instruction, and contains no literal `{SKILLS_CATALOG}` token.
- Given a lite chat prompt, when `_build_chat_prompt(text, lite=True)` renders, then the catalog block is absent.
- Running `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_awake.py -v` and `make lint` produces no failures.

### Open Questions

- Should the catalog list **all** human-facing commands or only a curated "high-signal" subset (the ~15-20 most useful) to keep the prompt lean? Defaulting to all human-facing commands grouped one-line-each unless you prefer a curated subset.
- Should suggestions be opt-in via a `config.yaml` flag (e.g. `chat.suggest_commands: true`), or always on? Defaulting to always-on since it's zero-cost and purely advisory — flag it if you'd rather gate it.

---

sukria:
Open questions answers : 
List only high value commands to save tokens 

Config optin : yes, this is the right way regarding our best practices.

@sukria-koan0 implement

PR #1805, opened 2026-06-07
fix(permissions): auto-create .claude/settings.json allowlist for instance/ files

### LINKED ISSUE #1798: Permission blocker: autonomous agents should not request approval for instance/* runtime files

tomgreen981111-cipher:
## What

Add `make permissions` and `make setup` auto-step that creates `.claude/settings.json` with a permission allowlist for `instance/**` files, so autonomous agents don't get blocked by approval prompts when writing to runtime state.

## Why

Closes #1798. Autonomous Kōan sessions (and any Claude Code agent running in the koan directory) were blocked by permission dialogs every time they wrote to `instance/missions.md`, `instance/outbox.md`, or journal/memory files. This is a fundamental contradiction of the autonomy model — the agent should own its own runtime state without per-operation approval dialogs.

The bootstrap catch-22: you can't create `.claude/settings.json` via the agent because the agent needs permission to create it. The fix is to generate the file from `make setup` (which the human runs), not from the agent itself.

## How

Three components:

1. **`koan/app/setup_claude_settings.py`** — idempotent installer with `install(project_root)` API. Writes `Write(instance/**)`, `Edit(instance/**)`, and common `Bash(git*)` / `Bash(gh*)` rules into `.claude/settings.json`. Preserves existing keys outside the `permissions` block.

2. **`Makefile`** — `.claude/settings.json` is now a dependency of `setup` (auto-created on first `make setup`). `make permissions` is a standalone target to refresh it.

3. **`koan/diagnostics/permissions_check.py`** — `/doctor` reports the allowlist state; `/doctor --fix` calls `install()` to create it if missing.

Also: `.gitignore` unignores `.claude/settings.json` (safe to commit since it's project config, not credentials), doctor handler registers the new check, troubleshooting docs explain the fix.

## Testing

- 17 new unit tests: `test_setup_claude_settings.py` + `test_permissions_check.py`
- All pass; pre-existing `test_us_timezone` failure is unrelated (missing `zoneinfo` timezone data on this host)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,11 @@
 # Kōan instance data (private to each user)
 instance
 .koan*
+
+# Claude Code local directory — track settings.json (permission allowlist) but
+# ignore session data, credentials, and machine-specific overrides.
 .claude
+!.claude/settings.json
 
 # Python
 __pycache__/
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 -include .env
 export
 
-.PHONY: install onboard setup start stop status restart
+.PHONY: install onboard setup permissions start stop status restart
 .PHONY: clean say migrate test test-skills test-strict coverage lint sync-instance rename-project release
 .PHONY: awake run errand-run errand-awake dashboard api api-token webhook
 .PHONY: ollama logs ssh-forward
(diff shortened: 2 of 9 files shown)


sukria:
False positive. My bad.
Closing this issue.
The real subject is mission triggering via chat messages.

PR #1812, opened 2026-06-07
fix(robustness): FD leak, atomic webhook, YAML triple-parse, redundant exception

tomgreen981111-cipher:
## What

Four targeted robustness/cleanup fixes across four files: FD leak, non-atomic file write, YAML over-parsing, and misleading exception tuple.

## Why

- **#1668** (`pid_manager.py`): `acquire_pidfile()` opened the PID file before the try block. A `KeyboardInterrupt` or other `BaseException` during startup would silently leak the fd, accumulating across crash-loop restarts.
- **#1671** (`github_webhook.py`): The check-notifications signal file was written with plain `open().write()`. A crash between open (which truncates) and write completes leaves an empty file. Every other shared-file write in the codebase uses `atomic_write()` — this was the only exception.
- **#1700** (`utils.py`): `resolve_project_path()` parsed `projects.yaml` up to 3× per call (step 1, step 6, `_resolve_via_fork_parent()`). On a 50-project config, the notification pipeline was doing triple disk+YAML work per mention. Now loaded once at step 1 and reused.
- **#1672** (`mission_runner.py`): `except (OSError, FileNotFoundError)` appeared 4× in the file. `FileNotFoundError` is a subclass of `OSError` — listing it explicitly implies they're distinct exception types, which confuses future contributors narrowing the catch.

Also fixes a pre-existing PERF401 lint violation in `recurring.py` (append loop → extend).

## How

- `pid_manager.py`: added `except BaseException: fh.close(); raise` after the existing `except (OSError, BlockingIOError)` clause.
- `github_webhook.py`: replaced plain `open().write()` with `atomic_write()` from `app.utils`.
- `utils.py`: `_projects_config` local variable captures the load at step 1; step 6 reads it directly; `_resolve_via_fork_parent()` gains an optional `config=` parameter so callers can pass it.
- `mission_runner.py`: `replace_all` edit to drop `FileNotFoundError` from all 4 tuples.

## Testing

Full suite: 15769 passed, 1 pre-existing failure (`test_us_timezone` — missing `tzdata` on this system, unrelated). Lint: clean.

### DIFF
--- a/koan/app/github_webhook.py
+++ b/koan/app/github_webhook.py
@@ -36,6 +36,7 @@ from typing import Optional, Set
 
 from app.github_config import DEFAULT_WEBHOOK_HOST, DEFAULT_WEBHOOK_PORT
 from app.signals import CHECK_NOTIFICATIONS_FILE
+from app.utils import atomic_write
 
 log = logging.getLogger(__name__)
 
@@ -151,8 +152,8 @@ def write_check_signal(koan_root: str) -> bool:
     """
     signal_path = os.path.join(str(koan_root), CHECK_NOTIFICATIONS_FILE)
     try:
-        with open(signal_path, "w") as f:
-            f.write(f"github webhook at {time.strftime('%H:%M:%S')}\n")
+        from pathlib import Path
+        atomic_write(Path(signal_path), f"github webhook at {time.strftime('%H:%M:%S')}\n")
         return True
     except OSError as e:
         log.warning("Webhook: failed to write check-notifications signal: %s", e)
--- a/koan/app/mission_runner.py
+++ b/koan/app/mission_runner.py
@@ -389,7 +389,7 @@ def _read_pending_content(instance_dir: str) -> str:
     pending_path = Path(instance_dir) / "journal" / "pending.md"
     try:
         return pending_path.read_text()
-    except (OSError, FileNotFoundError):
+    except OSError:
         return ""
 
 
@@ -412,7 +412,7 @@ def _read_stdout_summary(stdout_file: str, max_chars: int = 2000) -> str:
             return ""
         text = parse_claude_output(raw)
         return text[:max_chars] if text else ""
-    except (OSError, FileNotFoundError):
+    except OSError:
         return ""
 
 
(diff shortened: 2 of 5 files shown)


BabyKoan:
## PR Review — fix(robustness): FD leak, atomic webhook, YAML triple-parse, redundant exception

Four targeted robustness fixes, all sound. Merge-ready.

- **FD leak fix (pid_manager.py)**: Correct use of `except BaseException` to catch KeyboardInterrupt/SystemExit before fd leak. This is one of the few legitimate BaseException use cases.
- **Atomic webhook write (github_webhook.py)**: Properly eliminates race condition where crash between open() truncation and write() leaves empty signal file.
- **YAML triple-parse (utils.py)**: Clean optimization. Config loaded once, passed through call chain. Zero behavior change.
- **Redundant exception (mission_runner.py)**: Correctly removes `FileNotFoundError` from 4 tuples since it's an OSError subclass since Python 3.3.
- **PERF401 (recurring.py)**: Properly addresses lint violation with extend + generator.

---

### 🟢 Suggestions

<details>
<summary>
<b>1. Import placement style</b> (`koan/app/github_webhook.py`, L155-156)
</summary>

The `from pathlib import Path` import is placed inside the try block. While functionally correct, this is unconventional. Module-level imports are the project norm.

Consider moving to line 39 alongside the `atomic_write` import:
```python
from app.utils import atomic_write
from pathlib import Path
```

Then simplify line 155 to just:
```python
atomic_write(Path(signal_path), ...)
```

```
        from pathlib import Path
        atomic_write(Path(signal_path), f"github webhook at {time.strftime('%H:%M:%S')}\n")
```

</details>

---

### Checklist

- [x] No hardcoded secrets
- [x] Input validation at boundaries
- [x] Error handling - no bare except
- [x] Error handling - resource cleanup
- [x] No unsafe deserialization
- [x] Performance - no N+1 or repeated I/O
- [x] Python - mutable defaults
- [x] Python - resource management (with statement)


---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Ollama-launch · model qwen3.5:cloud)_ `HEAD=2e083fb` `1 min 19s`

BabyKoan (approved):
No blocking issues found.

atoomic (approved):

PR #1925, opened 2026-06-13
fix(missions): flush abandoned in-progress missions to Failed, not Done

exciton:
## Problem
When `start_mission()` finds stale In Progress missions (sanity enforcement), it was silently
moving them to Done with a ✅ marker — creating false history that the work completed
successfully. This fires when `recover.py` misses a mission (complex mission blocks, import errors).

## Changes
- Renamed `_move_in_progress_to_done()` → `_flush_abandoned_in_progress()` for clarity
- Changed the marker to ❌ with a `[flushed]` tag inserted into the Failed section
- Creates the Failed section if it doesn't already exist
- Updated all test assertions to check `sections["failed"]` instead of `sections["done"]`

## Test
All 415 existing tests pass (1 pre-existing root-permission test excluded).

### DIFF
--- a/koan/app/missions.py
+++ b/koan/app/missions.py
@@ -1089,11 +1089,16 @@ def _move_pending_to_section(
 
 
 def _flush_in_progress_to_done(content: str) -> str:
-    """Move all In Progress missions to Done.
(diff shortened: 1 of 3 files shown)


atoomic:
Thanks for fixing it

PR #1926, opened 2026-06-13
fix(recover): recover complex ### mission blocks from In Progress

exciton:
## Problem
Multi-step missions using the `### Header / - Step 1 / - Step 2` format were silently skipped
by crash recovery. They stayed in In Progress indefinitely and were never re-queued.

Additionally, `has_pending_journal` was computed once and applied to all in-progress missions.
With multiple stale missions, all were classified as "partial" even though pending.md was written
by exactly one interrupted run.

## Changes
- Replaced simple `in_complex_mission` flag with `_finalize_complex_block()` inner function
- Entire block (header line + sub-items) collected and classified together using the header as
  the mission key
- Recoverable block: all lines moved to Pending with `[r:N]` in the `### ` header
- Unrecoverable block: header line moved to Failed
- `journal_available` local flag initialized from `has_pending_journal`, set to `False` after
  first "partial" classification so subsequent missions are correctly classified as "dead"

## Test
70+ tests pass. `test_skip_complex_mission` renamed to `test_recover_complex_mission_block`
with updated assertions.

### DIFF
--- a/koan/app/recover.py
+++ b/koan/app/recover.py
@@ -289,10 +289,11 @@ def recover_missions(instance_dir: str, dry_run: bool = False) -> tuple:
                 _log_recovery_event(instance_dir, header, state, "escalated", attempts,
                                     has_checkpoint=has_checkpoint)
             else:
-                # Update counter in header, move entire block (header + sub-items) to Pending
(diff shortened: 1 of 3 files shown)


atoomic:
Thanks for the fix

PR #1936, opened 2026-06-14
docs(providers): OpenRouter via Claude CLI (CCR) guide

atoomic:
## What

Adds `docs/providers/openrouter.md` — how to run Koan's default Claude provider against **OpenRouter** models (a mix of Anthropic + cheaper non-Anthropic models) **without any code change**, using a local **CCR (claude-code-router)** server to repair tool-use/streaming so non-Anthropic models work in `claude -p` print mode.

## Why

Koan always invokes `claude -p --output-format json`, which demands strict Anthropic-style streaming + tool-use semantics. Many non-Anthropic OpenRouter models don't implement those faithfully, so pointing the CLI straight at OpenRouter's endpoint breaks `-p` missions. CCR's `tooluse`/`enhancetool` transformers + SSE rewriting fix this. (y-router is archived; claude-relay-service is an account pooler, not a model router.)

The integration needs **no Koan code**: it reuses existing hooks — `KOAN_CLAUDE_CLI_PATH`, `os.environ` inheritance into the CLI subprocess, and per-project `models:` passed verbatim as `--model`. Mode-pinning reuses the existing `usage.unlimited_quota` switch.

## Validated live

Against a running CCR instance, in the exact `claude -p` mode Koan uses:

| Test | `minimax/minimax-m3` | `qwen/qwen3.7-plus` |
|------|----------------------|---------------------|
| Print + JSON | ✅ `is_error:false`, `subtype:success` | ✅ same |
| Tool use (Bash) | ✅ `num_turns:2`, file actually written | ✅ same |

Also verified CCR routing: `--model openrouter,<slug>` is honored verbatim; plain tier names (`sonnet`) are hijacked by CCR's Router table. Documented in a "How model selection works" section.

## Changes

- **New:** `docs/providers/openrouter.md` (why CCR, architecture, install + launchd/systemd service, wrapper + `.env`, per-project model mapping, model-selection mechanics, `unlimited_quota` mode-pin, caveats, verification).
- Cross-links: `docs/README.md`, user-manual provider table, `docs/providers/claude.md`.
- `instance.example/config.yaml`: note `unlimited_quota` as the pay-per-token switch.

Docs-only — no Python touched.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,7 +13,7 @@ update the relevant docs in the same change.
 - [User Manual](users/user-manual.md) - daily use, workflows, and command guide.
 - [Onboarding](users/onboarding.md) - first-run setup and configuration flow.
 - [Skills Reference](users/skills.md) - built-in command reference.
-- [Provider Setup](providers/) - Claude, Cline, Codex, Copilot, local, and Ollama Launch providers.
+- [Provider Setup](providers/) - Claude, Cline, Codex, Copilot, local, and Ollama Launch providers; plus [OpenRouter via the Claude CLI](providers/openrouter.md).
 - [Messaging Setup](messaging/) - Telegram, Slack, Matrix, Discord, GitHub, and Jira.
 - [Troubleshooting](operations/troubleshooting.md) - common issues and how to fix them.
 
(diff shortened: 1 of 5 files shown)


sukria:
Looks great! I'll give it a try. Is open router a paying service? This might be documented as well. 

Also why not ship in the document a working example for relevant config entries?

PR #1954, opened 2026-06-15
refactor(run): decouple missions.md history pruning from finalization…

exciton:
## Problem
`prune_completed_sections()` ran inside the same locked read-modify-write that moves a mission
to Done/Failed (`_update_mission_in_file`). A pruning bug or misconfiguration could silently
mutate or corrupt history during the finalization write — two separate concerns coupled into
one transform, with the finalization result depending on pruning succeeding.

## Changes
- Extracted pruning into `_prune_missions_history()`: a standalone, locked, best-effort step
  run *after* the move commits. The finalization write now does only the mission move.
- The helper uses the missions lock (via `modify_missions_file`) so it cannot race the bridge
  inserting new missions — unlike the startup-time `startup_manager.prune_missions_done()`,
  which is safe unlocked only because nothing else writes during startup.
- A pruning error is logged and swallowed, leaving the committed move intact.
- Found-status now comes directly from `complete_mission_checked` / `fail_mission_checked`
  (the explicit `found` flag adopted from `main`) rather than from comparing file content
  before and after the write. With pruning decoupled, the move transform is the only writer,
  so an absent mission leaves the file untouched and can never be misreported as moved.

Pruning still runs per-finalization (missions.md stays bounded during long sessions), but as a
separate locked write — the coupling, not the cadence, was the problem.

## Test
`TestPruneDecoupledFromFinalization`:
- `test_finalization_triggers_history_prune` — oversized Failed section still trimmed to
  failed_keep after a completion
- `test_prune_failure_does_not_break_finalization` — a pruning RuntimeError leaves the move
  intact (returns True, Done updated, history uncorrupted)
- `test_prune_helper_is_noop_below_threshold` — small history left untouched
- `test_not_found_returns_false_and_skips_prune` — an absent mission reports False and, because
  pruning is decoupled, leaves an oversized history untouched

## Merge with main
Merged `origin/main` (72 commits). The only real conflict was in `_update_mission_in_file`:
`main` had independently hardened the same no-op-misreport bug by switching to the `_checked`
move functions, while this branch fixed it by decoupling pruning. The resolution keeps both —
`main`'s explicit `found` detection plus this branch's decoupled `_prune_missions_history`.
Also fixed a pre-existing `test_awake.py` failure on `main` (a `fake_stat` mock that didn't
accept Python 3.12's `follow_symlinks` kwarg).

### DIFF
--- a/koan/app/run.py
+++ b/koan/app/run.py
@@ -1815,9 +1815,14 @@ def _update_mission_in_file(
     the mission text could not be matched in Pending/In Progress). A False
     return means the mission is still in the queue and will be re-picked —
     callers should surface this rather than let it loop silently.
+
+    History trimming is intentionally NOT done here. Pruning old
+    Done/Failed entries is a separate maintenance concern run as its own
+    locked step (:func:`_prune_missions_history`) after the move commits, so
+    a pruning bug can never corrupt or roll back the finalization itself.
     """
     try:
-        from app.missions import complete_mission, fail_mission, prune_completed_sections
+        from app.missions import complete_mission, fail_mission
         from app.utils import modify_missions_file
         missions_path = Path(instance, "missions.md")
         if not missions_path.exists():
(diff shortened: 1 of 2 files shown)


atoomic:
@exciton needs a rebase

exciton:
> @exciton needs a rebase

Merged in main and fixed conflicts.

atoomic:
Thanks @exciton for your collaboration! 
I hope you get the best of the project

exciton:
> Thanks @exciton for your collaboration! 
> I hope you get the best of the project 
> 

You're very welcome!

PR #1961, opened 2026-06-15
refactor(retry): consolidate crash and stagnation counters; preserve in Failed state

exciton:
## Problem

Two independent retry systems accumulated silently with no shared ceiling:
- `[r:N]` tags embedded in mission text (missions.md) for crash-recovery, hardcoded max 3
- `.stagnation-retries.json` for stagnation retries, max configurable

A mission could cycle between stagnating and crashing indefinitely because:
1. Neither counter knew about the other
2. Any non-stagnation exit cleared the stagnation counter, resetting cross-system progress
3. Counters were cleared immediately on escalation to Failed, so the human couldn't see why

## Changes

**Commit 1 — cross-system ceiling (`total_attempts`):**
- Add `total_attempts` field to tracker entries; incremented by both stagnation requeues and
  crash-recovery on requeue
- New `max_total_retries` config key (default 0 = disabled) in `get_stagnation_config()` —
  single operator knob across both systems
- `clear_retry_count(clear_total=False)` preserves `total_attempts` across crash cycles
- Both `classify_mission_state()` and `_finalize_mission` check combined cap

**Commit 2 — unified storage (remove `[r:N]` from missions.md):**
- Rename `.stagnation-retries.json` → `.mission-retries.json` with auto-migration
- Add `crash_count` field alongside existing stagnation `count`
- New `get_crash_count()` / `increment_crash_count()` API; `increment_crash_count()` also
  increments `total_attempts`
- New `max_crash_retries` config key (default 3) replaces hardcoded `MAX_RECOVERY_ATTEMPTS`
- `classify_mission_state()` takes `crash_count: int` instead of parsing `[r:N]` from text
- Remove `MAX_RECOVERY_ATTEMPTS`, `_get_recovery_attempts`, `_set_recovery_attempts` from
  `recover.py`; keep `_strip_recovery_counter()` for backward-compat cleanup of old tags
- Backward compat: legacy `[r:N]` tags in existing missions.md are read during recovery and,
  when higher than the tracker's `crash_count`, seeded into the tracker (via `seed_crash_count`)
  so the migrated count survives the recovery cycle; the tag itself is stripped on next write

**Commit 3 — counter lifetime (preserve in Failed; clear on human retry):**
- Counter is NOT cleared when stagnation cap is hit or mission escalated to Failed
- Counter is cleared only when `_start_mission_in_file()` detects a cap was previously hit
  (`stag_count >= max_retry` OR `crash_count >= max_crash_retries` OR `total >= max_total`),
  signalling a deliberate human retry
- Ongoing stagnation-retry requeus (count < cap) keep their counter intact
- New `_clear_if_cap_hit()` helper encapsulates the conditional clear logic

**Commit 4 — PR review fixes:**
- `_clear_if_cap_hit` reads the tracker once via `get_retry_info()` and bails early when the
  mission has no entry — avoids a config load + 3 reads on every new-mission start (hot path).
  Config keys are read directly (no `.get()` fallbacks that could drift from the centralized
  `get_stagnation_config()` defaults).
- `total_cap` cause tag now embeds the counts: `stagnation:{pattern}:total_cap({total}/{max_total})`,
  making the Failed entry self-documenting.
- `classify_mission_state()` is now keyword-only, closing the silent positional-breakage footgun
  introduced by the `str → int` first-argument change.
- Legacy `[r:N]` seeding (see Commit 2 backward-compat note above).
- Documented the intentional counter-preservation-in-Failed behaviour and its bounded
  orphan-entry limitation in the `.mission-retries.json` tracker module comment.

**Commit 5 — PR review fixes (round 2):**
- **Project-scoped cap detection.** `_clear_if_cap_hit` now reads the *per-project* stagnation
  config (`project_name` threaded through `_start_mission_in_file` → `_clear_if_cap_hit`).
  It previously used global defaults while `_finalize_mission` used the project's, so a human
  retry on a project with tighter caps was silently re-escalated to Failed.
- **Cap-clear failures surface instead of being swallowed.** `_clear_if_cap_hit` returns a
  bool and narrows its catch to `ImportError`; schema mismatches in `get_retry_info` /
  `get_stagnation_config` now propagate. A cap-hit clear failure produces a prominent caller
  warning ("mission may re-escalate to Failed on next finalize") rather than a buried log line.
  Because the In-Progress transition already succeeded before this runs, the caller guards the
  call so a failure cannot abort an already-started mission.
- **Tracker import failure no longer disables escalation.** `recover.py` splits config load
  (safe to fall back to defaults) from tracker-function imports (the escalation safety net).
  When the tracker is unavailable it falls back to the legacy inline `[r:N]` counter **and
  persists the incremented count** (new `_set_recovery_counter()` helper) in both the simple
  and complex-block paths — so a repeatedly crashing mission still escalates instead of looping
  Pending→crash→Pending forever with a count that resets to 0 each cycle.

## Counter lifetime

| Event | Counter action |
|---|---|
| Crash → Failed | preserve all |
| Stagnation cap hit → Failed | preserve all |
| Escalated unrecoverable → Failed | preserve all |
| Stagnation retry requeue (count < cap) | preserve (cap still needs to fire) |
| Mission success | full clear |
| `start_mission()` with cap-hit counter | full clear (human deliberate retry) |

## Test

Full suite passes (2 pre-existing environment failures on the base commit, unrelated to this
change: a `fake_stat()`/`follow_symlinks` mock bug in `test_awake.py` and a flaky timeout in
`test_private_security_audit.py`).

New/updated: `TestCrashCount`, `TestRetryTracker`, `TestTotalAttempts`, `TestUnifiedRetryCap`,
`TestClassifyMissionState`, `TestMigrationBackwardCompat`, plus `TestClearIfCapHit`
(project-scoped cap clearing, no-op early return, clear-failure → `False`) and
`TestDegradedTrackerFallback` (inline counter persist / increment / escalate-at-cap).

**Commit 6 — PR review fixes (round 3):**
- **Consistent direct config access.** `_finalize_mission` now reads
  `max_retry_on_stagnation` / `max_total_retries` via direct key access
  (`cfg["..."]`) instead of `cfg.get(..., 0)`, matching `_clear_if_cap_hit`.
  `get_stagnation_config()` guarantees every key, so a missing key is a real
  schema bug that should surface loudly rather than silently defaulting — and
  the `0` fallback on `max_retry_on_stagnation` was semantically wrong (0
  disables retries; the actual default is 3).
- **Cap-clear errors logged at error level.** The caller's guard around
  `_clear_if_cap_hit` now logs propagated schema errors at error level with a
  full `traceback.format_exc()` instead of a truncated warning, so genuine bugs
  (e.g. a missing `get_retry_info` key) are not easy to miss in production.

### DIFF
--- a/koan/app/config.py
+++ b/koan/app/config.py
@@ -1125,6 +1125,8 @@ def get_stagnation_config(project_name: str = "") -> dict:
             per-system limits still apply). Default 0 (disabled). When set,
             provides a single operator knob to bound the total number of
             automatic retry attempts regardless of which system triggered them.
+        max_crash_retries (int): maximum crash-recovery attempts before
+            escalating a mission to Failed. Must be >= 1. Default 3.
 
     Per-project overrides via ``projects.yaml`` ``stagnation:`` take
     precedence. Setting ``enabled: false`` at project level disables the
@@ -1135,7 +1137,7 @@ def get_stagnation_config(project_name: str = "") -> dict:
         project_name: Optional project name for per-project overrides.
 
     Returns:
-        Dict with the resolved values — always contains all six keys.
+        Dict with the resolved values — always contains all seven keys.
     """
     defaults = {
         "enabled": True,
@@ -1144,6 +1146,7 @@ def get_stagnation_config(project_name: str = "") -> dict:
         "sample_lines": 50,
         "max_retry_on_stagnation": 3,
         "max_total_retries": 0,
+        "max_crash_retries": 3,
     }
     config = _load_config()
     base = config.get("stagnation", {})
@@ -1173,6 +1176,10 @@ def get_stagnation_config(project_name: str = "") -> dict:
     if max_total < 0:
         max_total = 0
 
+    max_crash = _safe_int(merged.get("max_crash_retries"), defaults["max_crash_retries"])
+    if max_crash < 1:
+        max_crash = 1
+
     return {
         "enabled": bool(merged.get("enabled", defaults["enabled"])),
         "check_interval_seconds": max(
@@ -1182,6 +1189,7 @@ def get_stagnation_config(project_name: str = "") -> dict:
         "sample_lines": max(1, _safe_int(merged.get("sample_lines"), defaults["sample_lines"])),
         "max_retry_on_stagnation": max_retry,
         "max_total_retries": max_total,
+        "max_crash_retries": max_crash,
     }
 
 
--- a/koan/app/config.py
+++ b/koan/app/config.py
@@ -1119,6 +1119,12 @@ def get_stagnation_config(project_name: str = "") -> dict:
             is re-queued before being marked Failed. ``0`` disables the
             retry loop entirely (mission is failed on the first stagnation).
             Default 3.
+        max_total_retries (int): ceiling on combined retry attempts across
+            both stagnation requeues and crash-recovery requeues for the same
+            logical mission. ``0`` disables the combined cap (independent
+            per-system limits still apply). Default 0 (disabled). When set,
+            provides a single operator knob to bound the total number of
+            automatic retry attempts regardless of which system triggered them.
 
     Per-project overrides via ``projects.yaml`` ``stagnation:`` take
     precedence. Setting ``enabled: false`` at project level disables the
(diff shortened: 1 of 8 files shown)


atoomic:
@exciton view issues above and conflicts, need a rebase

exciton:
> @exciton view issues above and conflicts, need a rebase

Merged in main, resolved above issues and fixed conflicts

atoomic (approved):

PR #2083, opened 2026-06-24
fix: stop infinite mission re-pick loop on [complexity:X] tags (#2082)

### LINKED ISSUE #2082: Infinite mission re-pick loop: start_mission() fails to move Pending items carrying a [complexity:X] tag
## Symptom

The agent loop enters an **infinite re-pick loop** on a single mission and makes no progress: the same mission is picked every iteration, the run aborts, and the loop restarts immediately. No GitHub reply, no other mission ever runs. Observed twice in a row — once on a `/plan` mission and once on an `/implement` mission (same issue #2081), identical behavior.

Log signature (repeats every ~3s):

```
[mission]  Mission picked: [koan] /plan .../issues/2081 📬 [complexity:simple] ⏳(20...
[git]      Ready on main from upstream
[warning]  Mission transition unconfirmed — '/plan .../issues/2081 📬 [com' not found in
           In Progress after start_mission(). Possible text normalisation mismatch or race condition.
[warning]  start_mission transition failed for '/plan .../issues/2081 📬 [com' — mission may
           still be in Pending; aborting this run to avoid duplicate execution.
=== Run 1/10 ... ===   (loops forever)
```

## Root cause

`koan/app/missions.py::_remove_item_by_text()` strips the `[complexity:X]` tag from each **candidate line** before comparing (`comparable = _COMPLEXITY_TAG_RE.sub("", stripped)`), but it does **not** strip the same tag from the **needle** (`line_needle`).

When the picked `mission_title` carries a `[complexity:simple]` tag (which it now does), the needle looks like:

```
... 📬 [complexity:simple] ⏳(...)
```

while the candidate `comparable` has the tag removed:

```
... 📬 ⏳(...)
```

So `line_needle in comparable` is `False` → the Pending item is never removed → `start_mission()` early-returns content unchanged → the post-transition confirmation in `run.py` fails ("transition unconfirmed") → the run aborts to avoid a duplicate → the picker re-selects the very same mission on the next iteration → infinite loop. The mission is never moved out of Pending, so quota burns with zero progress.

## Reproduction

With the real `instance/missions.md` (mission line carrying a `[complexity:simple]` tag):

```python
title = '[project:koan] /plan .../issues/2081 📬 [complexity:simple] ⏳(2026-06-24T12:23) ▶(2026-06-24T12:24)'
after = missions.start_mission(content, title)
assert after != content   # FAILS — mission not moved
```

Stripping `[complexity:X]` from the needle (the same way `comparable` already strips it) makes the match succeed.

## Proposed fix

In `_remove_item_by_text()`, apply `_COMPLEXITY_TAG_RE.sub("", ...)` to `line_needle` as well, so needle and candidate are normalized identically before the substring check.

```python
line_needle = re.sub(r"\s+", " ", _COMPLEXITY_TAG_RE.sub("", line_needle))
```

This mirrors the existing comment in the function (the whitespace-collapse fix) which already warns that a needle/candidate mismatch causes exactly this "re-dispatches on every loop iteration forever" symptom.

## Impact

- High: any mission whose picked title retains a `[complexity:X]` tag gets stuck in an infinite re-pick loop, blocking the entire queue and silently burning quota. `/plan` and `/implement` both hit it.

## Suggested tests

- `_remove_item_by_text()` / `start_mission()` matches a Pending line when the needle still contains `[complexity:simple]` but the file line's tag position/spacing differs.
- Regression: full pick → `start_mission` → confirmation round-trip succeeds for a mission carrying a complexity tag.

sukria:
## Problem

The agent loop entered an **infinite re-pick loop** on a single mission and made zero progress: the same `/plan` (then `/implement`) mission was picked every iteration, the run aborted with `Mission transition unconfirmed`, and the loop restarted immediately. No GitHub reply, no other mission ever ran, quota burned. Observed twice in a row (issue #2082).

## Root cause

`koan/app/missions.py::_remove_item_by_text()` strips the `[complexity:X]` tag from each **candidate** Pending line before comparing (`comparable = _COMPLEXITY_TAG_RE.sub("", stripped)`), but it did **not** strip the same tag from the **needle**.

When the picker handed `start_mission()` a title still carrying `[complexity:simple]`, the needle (`… 📬 [complexity:simple] ⏳…`) could never be a substring of the tag-stripped candidate (`… 📬 ⏳…`). The Pending item was never removed → `start_mission()` early-returned unchanged → `run.py`'s post-transition confirmation failed → the run aborted → the picker re-selected the same mission → ∞.

## Fix

Normalise the needle identically to the candidate (strip `[complexity:X]`, then collapse whitespace) before the substring check.

- **Self-heals** the stuck local state: a Pending line left with a stale `▶` started marker (crash mid-transition) can now be matched and moved instead of spun on forever — no manual `missions.md` edit needed.
- **Prevents recurrence**: needle and candidate now go through the same normalisation path.

## Tests

Two regression tests in `TestStartMission` reproduce the exact scenario (picked title with a `[complexity:X]` tag, and a Pending line carrying a stale `▶` marker). Both fail on `main`, pass with the fix. Full `test_missions*` / `test_pick_mission` / `test_complexity_tags` suites green; `make lint` clean.

Fixes #2082

### DIFF
--- a/koan/app/missions.py
+++ b/koan/app/missions.py
@@ -1043,12 +1043,16 @@ def _remove_item_by_text(
         (ln.strip() for ln in needle.splitlines() if ln.strip()),
         needle,
     )
-    # Collapse internal whitespace runs in the needle so it matches the
-    # collapsed `comparable` below. Without this, a mission whose text
-    # contains double spaces (e.g. inline /plan context) can never match —
-    # the runner succeeds but the mission is never moved out of Pending,
-    # so it re-dispatches on every loop iteration forever.
-    line_needle = re.sub(r"\s+", " ", line_needle)
+    # Normalise the needle the SAME way `comparable` is normalised below:
+    # strip any [complexity:X] tag, then collapse whitespace runs. The picker
+    # may hand us a title that still carries the `[complexity:simple]` tag
+    # while the stored line has had it stripped (or vice-versa); without
+    # normalising both sides identically the substring check never matches,
+    # the runner succeeds but the mission is never moved out of Pending, and
+    # it re-dispatches on every loop iteration forever (the infinite re-pick
+    # loop). The same hazard exists for missions whose text contains double
+    # spaces (e.g. inline /plan context).
+    line_needle = re.sub(r"\s+", " ", _COMPLEXITY_TAG_RE.sub("", line_needle))
 
     lines = content.splitlines()
     boundaries = find_section_boundaries(lines)
--- a/koan/tests/test_missions.py
+++ b/koan/tests/test_missions.py
@@ -1970,6 +1970,55 @@ class TestStartMission:
         assert "Already running task" in failed_text
         assert "[flushed]" in failed_text
 
+    def test_picked_title_with_complexity_tag_matches(self):
+        """Regression: infinite re-pick loop (issue #2082).
+
+        The picker can hand start_mission a title that still carries a
+        ``[complexity:X]`` tag while the stored Pending line has had its tag
+        position/spacing normalised. The needle and candidate must be
+        normalised identically, otherwise the mission is never moved out of
+        Pending and the agent loop re-dispatches it forever.
+        """
+        content = (
+            "# Missions\n\n"
+            "## Pending\n\n"
+            "- [project:koan] /plan https://example.com/issues/1 \U0001f4ec "
+            "[complexity:simple] ⏳(2026-06-24T12:23)\n\n"
+            "## In Progress\n\n"
+            "## Done\n"
+        )
+        # Title as returned by the picker — retains the complexity tag.
+        title = (
+            "[project:koan] /plan https://example.com/issues/1 \U0001f4ec "
+            "[complexity:simple] ⏳(2026-06-24T12:23)"
+        )
+        result = start_mission(content, title)
+        sections = parse_sections(result)
+        assert len(sections["pending"]) == 0
+        assert len(sections["in_progress"]) == 1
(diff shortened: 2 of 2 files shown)


sukria:
Works locally.

sukria:
@sukria-koan0 review

sukria:
@sukria-koan0 review

sukria-koan0:
## PR Review — fix: stop infinite mission re-pick loop on [complexity:X] tags (#2082)

Correct, minimal fix for the infinite re-pick loop. Merge-ready.

Strengths:
- Root cause precisely identified: `_remove_item_by_text` normalised the candidate but not the needle, so a `[complexity:X]`-tagged title could never substring-match a tag-stripped Pending line.
- The fix applies the *exact same* normalization pipeline to both sides (`_COMPLEXITY_TAG_RE.sub("")` then `\s+`→` `), eliminating the asymmetry rather than special-casing it — verified at missions.py:1055 vs :1070.
- Two behavioral regression tests reproduce the real scenarios (tagged picked title, and a stale `▶` marker for crash self-heal); they assert on observable section state, not source text.
- Comment rewritten to explain the symmetry invariant, which is the thing future editors must preserve.

No blocking issues. The expanded comment preserves the prior double-space rationale while documenting the new tag case.

---

---

### Checklist

- [x] Root cause addressed symmetrically
- [x] Regression tests verify behavior, not implementation
- [x] No new edge cases introduced


---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude · model opus)_ `HEAD=898a490` `20s`

sukria-koan0 (approved):
No blocking issues found.

PR #2088, opened 2026-06-24
fix(missions): stop zombie In Progress entries from start_mission TOCTOU (#2087)

### LINKED ISSUE #2087: Critical regression: fresh GitHub missions silently dropped as zombies (start_mission TOCTOU on [complexity:X] tag) + no Telegram on failure
## Summary

Every fresh GitHub-triggered mission (`/implement`, `/plan`, `/review` …) is **silently dropped** instead of executed. The mission is moved to `## In Progress` in `missions.md` but the run is aborted, leaving an orphaned "zombie" entry that is never executed and never re-picked. No Telegram notification is sent — the operator sees a mission marked *In Progress* / "running Xm" while nothing actually happens.

This is a **deterministic, critical** regression affecting the core mission loop.

## Reproduction

1. Trigger a mission from GitHub (e.g. `@bot implement` on an issue) so it lands in Pending with no `[complexity:X]` tag yet.
2. The loop picks it, classifies complexity (cache miss), and tries to start it.
3. Observed in `run.log`:
   ```
   ⚠️ Mission transition unconfirmed — '/implement …/issues/NNNN 📬'
      not found in In Progress after start_mission(). text normalisation mismatch or race.
   ⚠️ start_mission transition failed … aborting this run to avoid duplicate execution.
   ```
4. `missions.md` is left with a zombie:
   ```
   ## In Progress
   - [project:koan] /implement …/issues/NNNN 📬 [complexity:medium] ⏳(…) ▶(…)
   ```
   The mission never runs. The next picked mission flushes this one to `## Failed [flushed]` — also silently.

## Root cause (TOCTOU between two commits)

A time-of-check / time-of-use mismatch on the mission line, created by the interaction of two commits:

- **f7cc9422** (`feat: pre-classify mission complexity to route to cheapest capable model`, 2026-04-19) — introduced `tag_complexity_in_pending()` which injects `[complexity:X]` **in the middle of the line**, before the `⏳`/`▶` timestamp (`koan/app/missions.py`). The in-memory `mission_title` captured by the picker is **not** updated to reflect this on-disk rewrite.
- **921560bb** (`fix(run): verify start_mission transition and abort on mismatch`, 2026-06-12) — added a **strict** post-move confirmation in `_start_mission_in_file()` (`koan/app/run.py`): `clean_title in entry_text` (raw substring, whitespace-collapsed only). It **aborts** the run on mismatch. **This is the regression** — it turned the latent inconsistency planted by f7cc9422 into a hard failure.

The asymmetry: the *removal* step `_remove_item_by_text()` (missions.py) is complexity-tag-tolerant — it strips `[complexity:X]` before matching, so the move succeeds. The *confirmation* step (run.py) is **not** tolerant — and because the tag was injected between the mission body and the `⏳` timestamp, the captured `mission_title` (`… 📬 ⏳(…)`) is no longer a contiguous substring of the stored line (`… 📬 [complexity:medium] ⏳(…)`). Confirmation fails → abort → zombie.

It fires on **every cache miss** (fresh classification), i.e. every new GitHub mission — hence "totally unstable". Re-picked missions that already carry the tag confirm fine, which is why it looked intermittent.

### Secondary defect — silent failure

`koan/app/mission_executor.py` (~L920):
```python
if not _run._start_mission_in_file(instance, mission_title, project_name):
    log("warning", "… aborting this run to avoid duplicate execution.")
    return False   # no _notify, no _finalize, no Failed marking
```
The `🚀 Starting` Telegram notification comes *after* this check, so on abort the operator gets **no message at all**. Every abort/flush path is mute.

## Proposed fix

**A. Canonical shared matching.** Replace the raw substring confirmation in `run.py` with the same normalization used by removal (strip `[complexity:X]` + `[session:X]` + collapse whitespace), ideally via the existing `canonical_key()` (`missions.py`). One definition of "two lines are the same mission".

**B. Eliminate the TOCTOU.** In `mission_executor`, re-capture `mission_title` from the file after `tag_complexity_in_pending()` (or classify before capturing the title) so the in-memory title matches the on-disk line.

**C. Never fail silently → Telegram.** Any `start_mission` abort/flush path must emit a Telegram message `❌ [project] Mission failed: <reason>` and finalize the mission (or proceed when the line is canonically present, i.e. a false negative). No mission may die without an operator-visible message carrying the error.

## Acceptance criteria

- [ ] A fresh GitHub mission (cache-miss → mid-line `[complexity:X]` injected) starts and executes — zero zombies.
- [ ] `start_mission` confirmation uses the same canonical matching as removal.
- [ ] Any start/transition failure emits a Telegram notification containing the error reason.
- [ ] Regression tests: (1) cache-miss tag-injection → confirmation succeeds; (2) a genuine start failure → a Telegram error message is emitted.

Culprit commits: f7cc9422, 921560bb.

sukria:
Fixes #2087.

## Problem

Every fresh GitHub-triggered mission (`/implement`, `/plan`, `/review` …) was **silently dropped**: moved to `## In Progress` in `missions.md` but never executed, leaving an orphaned "zombie" entry that is never re-picked — and **no Telegram notification**. The operator saw a mission marked *In Progress* / "running Xm" while nothing happened. Deterministic, fired on every new GitHub mission.

## Root cause — TOCTOU between two commits

- **f7cc9422** (Apr 19) — `tag_complexity_in_pending()` injects `[complexity:X]` **in the middle** of the Pending line (before the `⏳` timestamp), after the picker captured `mission_title`. The in-memory title is never updated.
- **921560bb** (Jun 12) — added a **strict** post-move confirmation in `_start_mission_in_file()`: `clean_title in entry_text` (raw substring). It aborts on mismatch. This turned the latent inconsistency into a hard failure.

The removal step (`_remove_item_by_text`) is complexity-tag-tolerant so the move succeeds, but the confirmation was not — and because the tag is injected mid-line, the captured title (`… 📬 ⏳(…)`) is no longer a contiguous substring of the stored line (`… 📬 [complexity:X] ⏳(…)`). Confirmation fails → abort → zombie. Only fires on a cache miss (fresh classification), i.e. every new GitHub mission.

## Fix

- **`run.py`** — confirm the transition using the same canonical identity the removal step uses: `canonical_mission_key()` strips the `[complexity:X]` tag and lifecycle timestamps from **both** sides before the containment check. One definition of "two lines are the same mission".
- **`mission_executor.py`** — never fail silently: any unconfirmed start now emits a Telegram message `❌ [project] Mission not started: … Reason: …` so failures are operator-visible (the issue's explicit requirement).
- **tests** — regression reproducing the mid-line tag injection (confirmation must succeed, no zombie) + assert the abort path notifies Telegram.

## Verification

- New regression test fails on `main`, passes here.
- `koan/tests/test_run.py` + `test_complexity_tags.py`: **457 passed**.
- `make lint`: clean.

### DIFF
--- a/koan/app/mission_executor.py
+++ b/koan/app/mission_executor.py
@@ -918,8 +918,22 @@ def _run_iteration(
     original_mission_title = mission_title
     if mission_title:
         if not _run._start_mission_in_file(instance, mission_title, project_name):
+            reason = (
+                "could not confirm the Pending→In Progress transition "
+                "(mission not found in In Progress after start_mission)"
+            )
             log("warning", f"start_mission transition failed for '{mission_title[:60]}' — "
-                "mission may still be in Pending; aborting this run to avoid duplicate execution.")
+                f"{reason}; aborting this run to avoid duplicate execution.")
+            # Never fail silently: surface the abort on Telegram with the reason
+            # so the operator knows the mission did not run (see issue #2087).
+            try:
+                _run._notify(
+                    instance,
+                    f"❌ [{project_name}] Mission not started: {mission_title}\n"
+                    f"Reason: {reason}.",
+                )
+            except Exception as notify_err:  # notification must never mask the abort
+                log("error", f"Failed to notify start_mission abort: {notify_err}")
             return False
 
     # --- Create structured checkpoint for recovery ---
--- a/koan/app/run.py
+++ b/koan/app/run.py
@@ -2197,12 +2197,23 @@ def _start_mission_in_file(instance: str, mission_title: str, project_name: str 
 
         after = modify_missions_file(missions_path, _transform)
         in_progress = parse_sections(after).get("in_progress", [])
-        # Normalise for comparison: strip leading "- ", collapse whitespace
+        # Confirm the transition using the SAME canonical identity that the
+        # removal step (_remove_item_by_text) uses, not a raw substring. The
+        # complexity classifier rewrites the on-disk Pending line by injecting
+        # ``[complexity:X]`` *between* the mission body and the ⏳ timestamp
+        # (tag_complexity_in_pending) AFTER ``mission_title`` was captured. A
+        # naive ``mission_title in entry`` check then fails because the in-memory
+        # title (``… 📬 ⏳(…)``) is no longer a contiguous substring of the stored
+        # line (``… 📬 [complexity:X] ⏳(…)``), even though the move succeeded —
+        # leaving an orphaned "zombie" In Progress entry. canonical_mission_key()
+        # strips the complexity tag and lifecycle timestamps from both sides, so
+        # the same logical mission matches regardless of when the tag was added.
         import re
-        clean_title = re.sub(r"\s+", " ", mission_title.strip())
+        from app.missions import canonical_mission_key
+        clean_title = re.sub(r"\s+", " ", canonical_mission_key(mission_title))
         for entry in in_progress:
-            entry_text = re.sub(r"\s+", " ", entry.strip().removeprefix("- "))
-            if clean_title in entry_text:
+            entry_text = re.sub(r"\s+", " ", canonical_mission_key(entry))
+            if clean_title and clean_title in entry_text:
                 # Clear counter only if a cap was previously hit (human deliberate
                 # retry) — stagnation-retry requeus must keep their count intact
                 # so the stagnation cap check in _finalize_mission still fires.
--- a/koan/tests/test_run.py
+++ b/koan/tests/test_run.py
@@ -6408,6 +6408,43 @@ class TestStartMissionSanityFlushLog:
         assert "leftover stale" not in "\n".join(sections.get("failed", []))
 
 
+class TestStartMissionComplexityTagTOCTOU:
+    """Regression for #2087: a [complexity:X] tag injected into the Pending
+    line *after* the mission title was captured must not break the transition
+    confirmation. Before the fix, the raw-substring check failed because the
+    captured title (``… 📬 ⏳(…)``) was no longer a contiguous substring of the
+    stored line (``… 📬 [complexity:medium] ⏳(…)``), leaving a zombie entry.
+    """
+
+    def test_transition_confirmed_when_complexity_tag_injected_midline(self, tmp_path):
+        from app.run import _start_mission_in_file
+        from app.missions import parse_sections
+
+        missions = tmp_path / "instance" / "missions.md"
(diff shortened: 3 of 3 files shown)


sukria:
@sukria-koan0 review

sukria:
Fixes the issue locally.

sukria-koan0:
## PR Review — fix(missions): stop zombie In Progress entries from start_mission TOCTOU (#2087)

Tight, correct fix for the TOCTOU zombie-mission bug. Merge-ready.

Strengths:
- Root cause is precisely diagnosed and the fix reuses the existing `canonical_mission_key()` identity (verified at missions.py:203) — same normalization the removal step relies on, so confirmation and removal now agree by construction.
- The `if clean_title and ...` guard correctly prevents an empty canonical key from matching any entry.
- The Telegram abort notification wraps `_notify` in try/except so a notification failure can't mask the abort — and logs the error instead of swallowing it.
- Tests assert observable behavior: the regression test checks the mission actually lands in In Progress and leaves no Pending zombie; the abort test asserts the operator-facing notification fires with the mission name.

- One minor, pre-existing note: the substring-containment confirmation does not key on `project_name` (suggestion only, not a regression).

---

---

### Checklist

- [x] Root cause correctly addressed
- [x] No silent failure path (abort now notifies)
- [x] Error handling around notification
- [x] Tests verify behavior, not implementation
- [ ] Match logic robust across projects


---

## Silent Failure Analysis

<details>
<summary>
🟡 **MEDIUM** — silent no-op on empty key<br>
<a href="https://github.com/Anantys-oss/koan/blob/05918128004708db300e0344dcd2505a905498d3/koan/app/run.py#L2210-L2216">koan/app/run.py:2210-2216</a>
</summary>

**Risk**: If canonical_mission_key(mission_title) yields an empty string (e.g. a title consisting only of tags/timestamps), the new `if clean_title` guard makes the confirmation loop match nothing and silently skip the counter-clear, with no log explaining why the transition appeared unconfirmed.

```
clean_title = re.sub(r"\s+", " ", canonical_mission_key(mission_title))
for entry in in_progress:
    entry_text = re.sub(r"\s+", " ", canonical_mission_key(entry))
    if clean_title and clean_title in entry_text:
```

**Fix**: Log a warning when clean_title is empty so an unexpected empty canonical key is surfaced rather than silently treated as 'no match'.

</details>

---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude · model opus)_ `HEAD=0591812` `1 min 10s`

sukria-koan0 (approved):
No blocking issues found.

PR #2177, opened 2026-06-26
perf(run): stream skill-dispatch stdout to disk, not RAM (#2173)

### LINKED ISSUE #2173: MAJOR: stream Claude CLI stdout instead of buffering full transcript in RAM (11 GB spikes)
Part of #2170 (RAM footprint optimization). **This is the single biggest win for the 11 GB spikes.**

## Problem

In `run_claude_task` (`run.py`), every line of Claude CLI stdout is appended to an in-memory list `stdout_lines` (`run.py:2692`, `:2770`) and kept for the **entire** mission, then joined into one giant string (`run.py:2789`, `:2845`, `:2853`). A long, verbose mission can accumulate hundreds of MB to GBs; the final `"\n".join(...)` then creates a **second** full copy.

Crucially, the same loop **already streams every line to `pending.md` on disk** (`run.py:2772-2777`), so the full in-memory copy is redundant.

```python
stdout_lines = []                       # run.py:2692
for line in proc.stdout:                # run.py:2766
    stripped = line.rstrip("\n")
    stdout_lines.append(stripped)       # run.py:2770  <-- unbounded growth
    print(stripped)
    if pending_fh is not None:
        pending_fh.write(f"{stripped}\n")   # already on disk
...
skill_stdout = "\n".join(stdout_lines)  # run.py:2789  <-- second full copy
```

Only the **tail** is actually needed in memory (`tail_lines = stdout_lines[-20:]` at `run.py:2831`).

## Fix

1. Replace `stdout_lines: list` with `collections.deque(maxlen=N)` (e.g. 100 lines) — caps peak RAM from GBs to a few KB.
2. When the full transcript is genuinely needed for post-mission processing, read it back from `pending.md` / the stdout sidecar instead of holding it in RAM.
3. Audit other subprocess-output collection sites for the same pattern.

With `max_parallel_sessions > 1` this cost multiplies per concurrent mission (`_live_sessions`, `run.py:1509`) — fixing this is what removes the 11 GB peaks. (Related Railway config: pin `max_parallel_sessions: 1` — tracked in #2170.)

sukria-koan0:
## Summary

`_run_skill_mission` buffered every skill-runner stdout line into an unbounded Python list and then built a second full copy with `"\n".join(...)` — even though every line was already streamed to `pending.md` and the joined string already written to `stdout_file`. For verbose skill sessions this peaked at hundreds of MB, multiplied per concurrent session. This streams each line straight to disk during the loop, keeps only a 200-line `deque` tail in RAM, and reads the transcript back from disk once when a caller genuinely needs it.

Closes https://github.com/Anantys-oss/koan/issues/2173

## Changes

- `run.py`: add `_pump_skill_stdout()` + `_read_back_or_tail()`; rewrite `_run_skill_mission` to stream stdout to `stdout_file` + bounded tail; drop the `stdout_lines` list and all three `"\n".join()` copies; append usage-JSON to `stdout_file`; drop the now-redundant post-loop file write.
- `provider/__init__.py`: bound `raw_lines` (error/max-turns detection only) with `deque(maxlen=2000)`; document why `text_lines` (the return value) stays unbounded.
- Tests: `test_skill_stdout_streaming.py` (pump streams to both sinks, bounds tail, survives pending-write failure); `test_cli_exec.py` regression guarding the streaming result accumulator stays full.
- Docs: `daemon.md` "CLI stdout memory model" subsection.

## Test plan

- `pytest tests/test_skill_stdout_streaming.py tests/test_cli_exec.py tests/test_run.py` — pass
- `pytest -k "provider or skill or mission_exec"` — 4182 pass
- `make lint` — clean
- Confirmed `grep stdout_lines koan/app/run.py` returns nothing

---
_Generated by [Kōan](https://koan.anantys.com)_ _(claude · model claude-opus-4-8 · HEAD=0fe77ccd)_

### DIFF
--- a/docs/architecture/daemon.md
+++ b/docs/architecture/daemon.md
@@ -126,3 +126,27 @@ path because they depend on git prep and specialised post-mission handling.
 
 Single-slot installations (`max_parallel_sessions: 1`, the default) skip all
 parallel logic with zero overhead.
+
+## CLI stdout memory model
+
+Claude/skill CLI output can reach hundreds of MB per mission, multiplied per
+concurrent session when `max_parallel_sessions > 1`. To keep peak RAM bounded,
+no path holds the full transcript in a Python list while the mission runs:
+
+- **Main mission path** (`run_claude_task`): the child's stdout is wired
+  straight to a temp file (`subprocess.Popen(stdout=out_f)`); no Python-side
+  line list is ever built. Downstream consumers (`run_post_mission`, token
+  parsing, PR-URL extraction) read that file.
+- **Skill-dispatch path** (`_run_skill_mission`): each line is streamed via
+  `_pump_skill_stdout` to the same on-disk capture and to `pending.md` (for
+  `/live`) as it arrives; only a 200-line tail deque is kept in RAM for timeout
+  diagnostics. The full transcript is read back from disk once, at
+  end-of-mission, when a caller needs it (`_extract_pr_url`, the `— skipping`
+  check, error classification).
+- **Provider stream runner** (`run_command_streaming`): the error/max-turns
+  accumulator (`raw_lines`) is a bounded `deque`; the assistant-text
+  accumulator (`text_lines`) is the actual return value and is intentionally
+  unbounded so long sessions never silently lose output.
+
+Operators running on constrained hosts (e.g. Railway) should also pin
+`max_parallel_sessions: 1` so per-session cost is not multiplied.
--- a/koan/app/provider/__init__.py
+++ b/koan/app/provider/__init__.py
@@ -27,6 +27,7 @@ import re
 import subprocess
 import sys
 import tempfile
+from collections import deque
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
@@ -870,7 +871,13 @@ def run_command_streaming(
 
     from app.cli_exec import popen_cli
 
-    raw_lines: List[str] = []  # for error reporting (full transcript)
+    # raw_lines is scanned only for error/max-turns detection, never returned,
+    # so a bounded tail is safe and caps RAM on long provider streams. 2000 is
+    # deliberately generous so terminal _format_cli_error context and the
+    # non-stream-json max-turns regex fallback keep working.
+    raw_lines = deque(maxlen=2000)  # for error reporting (terminal lines)
+    # text_lines IS the fallback return value when no result event arrives —
+    # it must stay unbounded or long sessions would silently lose output.
     text_lines: List[str] = []  # fallback return value when no result event
     final_result: Optional[str] = None
     usage_snapshot: Optional[Dict[str, Any]] = None
--- a/koan/app/run.py
+++ b/koan/app/run.py
@@ -31,6 +31,7 @@ import tempfile
 import threading
 import time
 import traceback
+from collections import deque
 from pathlib import Path
 from typing import List, Optional, Tuple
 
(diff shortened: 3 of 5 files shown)


sukria:
@sukria-koan0 review

sukria-koan0:
## PR Review — perf(run): stream skill-dispatch stdout to disk, not RAM (#2173)

Solid, well-scoped memory fix — streams skill stdout to disk and keeps only a 200-line tail in RAM. Merge-ready.

Strengths:

- The `out_fh` lifecycle is handled correctly across all exit paths: closed-then-`None` before the watchdog-timeout raise (so the read-back sees flushed data), flush-only in the `except` branches with the close deferred to `finally`, and a single `out_fh is None` guard prevents double-close. No leak or double-close found.
- Reading the full transcript back from disk once at end-of-mission is justified — I confirmed the three consumers (`_extract_pr_url`, the `"— skipping"` check, and `_classify_and_handle_cli_error`) in mission_executor.py:139-219 genuinely need the whole transcript, not just a tail, so the bounded tail alone wouldn't suffice.
- The `provider/__init__.py` change correctly bounds only `raw_lines` (error-detection scratch) while leaving `text_lines` (the return value) unbounded, with a comment explaining why — and the `test_cli_exec.py` regression test locks that contract in.
- Docs (`daemon.md`) accurately describe all three paths; I verified the `run_claude_task` claim against run.py:345-348.

- Only nit: the new on-disk write drops explicit utf-8 encoding (suggestion below) — narrow edge, non-blocking.

---

### 🟢 Suggestions

<details>
<summary>
<b>1. Skill stdout file now uses locale default encoding (was explicit utf-8)</b><br>
<a href="https://github.com/Anantys-oss/koan/blob/0fe77ccd40d7d226595ce9dd3efa3aa99c1b91d5/koan/app/run.py#L2764">koan/app/run.py:2764</a>
</summary>

`out_fh = open(stdout_file, "w")` and `_read_back_or_tail`'s `read_text(errors="replace")` both rely on the locale default encoding, whereas the removed final write used explicit `utf-8` (`skill_stdout.encode('utf-8')`).

Why it matters:

- The main mission path (`run_claude_task`, run.py:345) also opens `stdout_file` in text mode, but it passes the handle straight to `Popen(stdout=out_f)` — the child writes bytes to the fd directly, bypassing Python's encoder. This new skill path instead does Python-side `out_fh.write(stripped)`, so it *is* subject to the process encoding.
- On a non-UTF-8 locale (e.g. `LC_ALL=C`/`POSIX`, common in minimal containers and cron), a non-ASCII transcript line would raise `UnicodeEncodeError` inside the **unguarded** `out_fh.write` in `_pump_skill_stdout`. Unlike the `pending_fh` write (wrapped in `try/except OSError`), this propagates out of the pump → `except Exception` → `exit_code=1`, turning an otherwise-successful skill mission into a failure. The old code accumulated lines in RAM and only encoded once at the end, inside a try that logged rather than failed the mission.

Fix: pass `encoding="utf-8"` to both the `open(stdout_file, "w")` and the `Path(stdout_file).read_text(...)` call so the round-trip is locale-independent and matches the prior explicit-utf-8 behavior.

```
out_fh = open(stdout_file, "w")  # default encoding; prior code encoded utf-8 explicitly
```

</details>

---

### Checklist

- [x] out_fh lifecycle correct across normal/timeout/exception paths (no leak, no double-close)
- [x] Full-transcript read-back justified by real consumers
- [x] Provider accumulator bounding preserves return-value completeness
- [ ] Encoding consistency on disk I/O — suggestion ＃1
- [x] Docs match code
- [x] Test coverage for streaming + bounding + write-failure


---

## Silent Failure Analysis

<details>
<summary>
🟡 **MEDIUM** — silent partial-data fallback on error<br>
<a href="https://github.com/Anantys-oss/koan/blob/0fe77ccd40d7d226595ce9dd3efa3aa99c1b91d5/koan/app/run.py#L2682-L2687">koan/app/run.py:2682-2687</a>
</summary>

**Risk**: If the on-disk transcript read fails, this silently returns only the last 200 lines (the tail deque) with no log, so error classification, _extract_pr_url, and the '— skipping' check operate on a truncated transcript — a quota/auth error or PR URL emitted early in a long session would be silently lost and the mission misclassified.

```
def _read_back_or_tail(stdout_file, tail):
    try:
        return Path(stdout_file).read_text(errors="replace")
    except OSError:
        return "\n".join(tail)
```

**Fix**: Log a warning (e.g. via log/debug_log) inside the except branch so the truncated-transcript fallback is visible, making downstream misclassification diagnosable.

</details>

---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude · model opus)_ `HEAD=0fe77cc` `3 min 43s`

sukria:
@sukria-koan0 rebase

fix the encoding suggestion → Fix: pass encoding="utf-8" to both the open(stdout_file, "w") and the Path(stdout_file).read_text(...) call so the round-trip is locale-independent and matches the prior explicit-utf-8 behavior.

sukria-koan0:

Branch `koan0/implement-2173` was rebased onto `main` and review feedback was applied.

### Changes applied

- Done. Both encoding fixes applied.
- ## Summary
- Added `encoding="utf-8"` to `open(stdout_file, "w")` in `_run_skill_mission` (run.py:2803), per reviewer suggestion #1. Prevents `UnicodeEncodeError` on non-UTF-8 locales (e.g. `LC_ALL=C`) in the unguarded `out_fh.write` inside `_pump_skill_stdout`, which would turn a successful skill mission into a failure.
- Added `encoding="utf-8"` to `Path(stdout_file).read_text(...)` in `_read_back_or_tail` (run.py:2682), making the disk round-trip locale-independent.
- Both restore the explicit UTF-8 behavior of the removed `skill_stdout.encode('utf-8')` write. Did not touch the MEDIUM silent-failure suggestion — human asked only for the encoding fix.

### Stats

```
5 files changed, 174 insertions(+), 28 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main touches skill-dispatch stdout streaming or the unbounded buffer in _run_skill_miss)
- Rebased `koan0/implement-2173` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/implement-2173` to origin

</details>

### CI status

---
_Automated by Kōan_

Issue #2174, opened 2026-06-26
Don't launch dashboard on KOAN_DEPLOY=railway unless KOAN_DASHBOARD_PWD is set


sukria:
Part of #2170 (RAM footprint optimization).

## Goal

On `KOAN_DEPLOY=railway`, do **not** launch the web dashboard unless `KOAN_DASHBOARD_PWD` is set. This both (a) keeps the dashboard from ever being exposed publicly without a passphrase, and (b) lets us measure the **pure Kōan worker footprint** (run + awake only) without the Flask dashboard process.

## Current state — two conflicting launch paths

1. **`koan/docker/dashboard-supervised.sh` already does this correctly**: it refuses to start (`exec sleep infinity`) when `KOAN_DEPLOY != railway` OR when `KOAN_DASHBOARD_PWD` is unset.
2. **But `start_all()` launches the dashboard independently** whenever `dashboard.enabled: true` in `config.yaml` (`pid_manager.py:638-641` → `_is_dashboard_enabled()` → `config.is_dashboard_enabled()`), with **no passphrase / no railway gate**.

On a fresh Railway restart the dashboard was observed running — i.e. the config-based path (2) starts it regardless of `KOAN_DASHBOARD_PWD`, bypassing the gate in (1).

## Ask

1. Make the config-driven launch path (`start_all` / `_is_dashboard_enabled`) honor the same gate: on `KOAN_DEPLOY=railway`, refuse to start the dashboard unless `KOAN_DASHBOARD_PWD` is set.
2. Single source of truth — ideally both paths call one helper (e.g. `railway.dashboard_allowed()`).
3. Outcome: with `KOAN_DASHBOARD_PWD` unset on Railway, only `run` + `awake` run, exposing the true minimal worker footprint (observed settled baseline ~459 MB).

sukria:
@sukria-koan0 implement

Issue #2172, opened 2026-06-26
ollama serve should never launch when provider != ollama (wasted RAM)


sukria:
Part of #2170 (RAM footprint optimization).

## Problem

`ollama serve` is being launched on a deploy that does **not** use the Ollama provider. `ollama serve` is the single largest idle memory consumer in the stack, so this is pure wasted footprint.

## Where

`start_all()` only starts ollama when `_needs_ollama(provider)` is true, and that is gated on `provider == "ollama"` (`pid_manager.py:577`). `provider` comes from `_detect_provider()` → `app.provider.get_provider_name()` (env `KOAN_CLI_PROVIDER`/`CLI_PROVIDER` > `config.yaml` > default).

So if `ollama serve` is running on a Claude deploy, either:
- provider resolution is returning `ollama` when it shouldn't (mis-set env/config), or
- another launch path starts ollama outside `start_all`'s gate, or
- a stale `ollama` pidfile/process survived a provider switch.

## Ask

1. Confirm the deployed provider resolves to `claude` (log the resolved provider name at startup so this is visible).
2. Ensure `ollama serve` is **never** started unless the resolved provider is exactly `ollama` — audit every launch path (`start_all`, `start_stack`, `make ollama`, docker entrypoint/supervisord) for an ungated ollama start.
3. On Railway / `KOAN_DEPLOY=railway`, refuse to start ollama entirely unless explicitly opted in, since the hosted profile should default to the Claude provider.

sukria:
@sukria-koan0 implement

Issue #2170, opened 2026-06-26
Optimize RAM footprint: ~2 GB idle baseline and 11 GB mission spikes (Railway cost)


sukria:
## Problem

Kōan's deployed RAM footprint is too high for cost-efficient Railway hosting.

Observed on the Railway memory dashboard (`koan-saas` service):

- **Idle baseline: ~1.3–2.0 GB** even when the agent does nothing (no mission running).
- **Spikes to ~11 GB** during mission execution.

Target: **minimum possible footprint when idle**, and bounded peaks during work.

## Investigation summary

Two failure modes:

1. **Idle baseline** is driven by the number of always-on Python processes and their imported libraries — not by a leak.
2. **11 GB spikes** are driven by buffering the entire Claude CLI subprocess output in memory for the lifetime of a mission.

---

### A. Idle baseline (~2 GB)

`make start` / `start_all()` launches up to **5 long-lived processes** (`pid_manager.py:261`):
`run`, `awake`, `ollama`, `dashboard`, `api`.

Each Python process pays a fixed import cost. The heavy import chains pulled in at startup:

- `flask` → werkzeug + jinja2 + click (dashboard **and** api)
- `waitress` (api WSGI server)
- `textual` → rich (TUI)
- `requests` → urllib3 + certifi

A bare Python process with `flask`+`requests`+`textual` resident is ~60–120 MB **per process**. With dashboard + api + run + awake all up, that alone is several hundred MB before any work — and `ollama serve` (if the Ollama provider is used) is the single biggest idle consumer by far.

**Recommendations (idle):**

1. **Do not run optional processes on Railway.** The dashboard and REST API should be opt-in, off by default in the deployed profile. Gate them so a minimal deploy runs only `run` + `awake`. (`pid_manager.start_all` lines 640–645 start dashboard/api when enabled — make sure the Railway `config.yaml` disables both.)
2. **Don't co-locate `ollama serve`** with the agent on a small Railway instance — it dwarfs everything else. Use the Claude provider in the deployed profile.
3. **Lower `api.threads`** from the default of `8` (`config.py:get_api_threads`) to `2` if the API is enabled at all — each waitress worker holds request buffers + a DB connection.
4. **Lazy-import heavy modules.** `flask`, `waitress`, `textual`, `slack-sdk` should be imported inside the functions that need them, not at module top-level, so the `run`/`awake` processes never pay for web/TUI libraries they don't use.

### B. Mission spikes (~11 GB) — the biggest single win

[...]

PR #2178, opened 2026-06-26
perf(memory_db): stream JSONL reindex/migration (item D of #2170)

sukria-koan0:
## What
Stream `log.jsonl` line-by-line when reindexing/migrating into the SQLite FTS5 index, instead of buffering the whole file in RAM.

## Why
Closes item **D** of #2170 (RAM optimization). The three MAJOR sub-issues (#2172 ollama gating, #2173 stdout deque, #2174 dashboard gating) are merged — they killed the 11 GB mission spikes and cut the idle baseline. Item D was the remaining mechanism flagged as "secondary": `_reindex_from_jsonl` and `migrate_jsonl_to_sqlite` each did `read_text()` + `splitlines()` + a full `entries` list, holding **three copies** of a file that grows unbounded over an instance's lifetime. Fires only on startup/migration, but can transiently spike RAM on a large instance.

## How
Extracted one shared `_stream_jsonl_into_entries()` helper that reads one line at a time and flushes a batched `executemany` every 500 rows, so peak memory stays flat regardless of log size. Both readers now share it (also dedups the identical 8-column tuple projection via `_row_from_obj`). Error contracts preserved: `OSError` on open → bail, `sqlite3.DatabaseError` on insert → warn-and-degrade.

## Testing
- `make lint` — clean.
- `make test` (memory_db): 50 passed, 1 skipped.
- New test `test_migrate_streams_across_batch_boundary_and_skips_malformed` writes 1007 rows + a malformed line, asserts all valid rows land (proves the batch-flush path indexes the full file, not just the last batch).

## Note on the broader question
Asked whether *significant* gains remain: no. A4 (lazy-import flask/waitress/textual/slack) is already effectively satisfied — measured `app.run` and `app.awake` import **zero** of those heavy libs (~23 MB RSS). This PR closes the last concrete item; remaining ideas (mtime-caching `missions.md`/journals) are micro-optimizations on already-small hot-path files.

---
🤖 Generated from analysis of #2170

### DIFF
--- a/docs/operations/rest-api.md
+++ b/docs/operations/rest-api.md
@@ -23,7 +23,7 @@ api:
   enabled: true       # Include API in managed processes (default: false)
   host: "127.0.0.1"  # Bind address (default: 127.0.0.1 — loopback only)
   port: 8420          # HTTP port (default: 8420)
-  threads: 8          # waitress worker threads (default: 8)
+  threads: 2          # waitress worker threads (default: 2)
   # token: ""         # Bearer token fallback (prefer KOAN_API_TOKEN env var)
 ```
 
--- a/docs/setup/railway.md
+++ b/docs/setup/railway.md
@@ -98,6 +98,19 @@ This gate is enforced on **both** launch paths through a single helper
 `config.yaml`). With the passphrase unset on Railway, only `run` + `awake`
 launch, exposing the minimal worker footprint.
 
+## REST API gate (`KOAN_API_TOKEN`)
+
+The optional REST API follows the same "no optional processes on Railway unless
+explicitly opted in" policy. Every extra long-lived process raises the
+container's idle RAM floor, so the config-driven launcher only starts the API on
+Railway when `KOAN_API_TOKEN` is set — the token is required for the API to
+function at all, so its presence is the explicit opt-in. This is enforced by
+`railway.api_allowed()` inside `pid_manager._is_api_enabled`. Off Railway the
+gate is inert and `api.enabled: true` in `config.yaml` alone decides.
+
+The API also defaults to **2** waitress worker threads (`api.threads`), down from
+8, to keep its resident footprint small on memory-constrained hosts.
+
 `make koan` either **attaches** to the already-running daemon (status/logs/
 dashboard), or runs the onboarding **wizard** on an empty volume. Because the
 volume is made writable at boot, config edits (`instance/config.yaml`,
--- a/instance.example/config.yaml
+++ b/instance.example/config.yaml
@@ -699,7 +699,7 @@ usage:
 #   enabled: false          # Include API in managed processes (default: false)
 #   host: "127.0.0.1"       # Bind address (default: 127.0.0.1)
 #   port: 8420              # HTTP port (default: 8420)
-#   threads: 8              # waitress worker threads (default: 8)
+#   threads: 2              # waitress worker threads (default: 2)
 #   # token: ""             # Bearer token fallback (prefer KOAN_API_TOKEN env var)
 
 # Automation suggestions — context-aware recurring task proposals
--- a/koan/app/config.py
+++ b/koan/app/config.py
@@ -792,15 +792,20 @@ def get_api_token() -> str:
 
 
 def get_api_threads() -> int:
-    """Return the number of waitress worker threads (default: 8).
+    """Return the number of waitress worker threads (default: 2).
 
-    Config key: api.threads (default: 8)
+    Lowered from 8 to 2: the REST API is a low-traffic control plane, and
+    each idle waitress thread holds a stack + thread-local arena. Two threads
+    keep the resident footprint small on memory-constrained hosts (Railway)
+    while still serving concurrent requests. Operators can raise it via config.
+
+    Config key: api.threads (default: 2)
     """
     config = _load_config()
     api_cfg = config.get("api", {})
     if isinstance(api_cfg, dict):
-        return _safe_int(api_cfg.get("threads", 8), 8)
-    return 8
+        return _safe_int(api_cfg.get("threads", 2), 2)
+    return 2
 
 
 def get_cli_output_journal() -> bool:
--- a/koan/app/memory_db.py
+++ b/koan/app/memory_db.py
@@ -133,12 +133,18 @@ def _reindex_from_jsonl(conn: sqlite3.Connection, instance: str) -> None:
     if not log_path.exists():
         return
     try:
-        inserted, _ = _stream_jsonl_into_entries(conn, log_path)
+        inserted, skipped = _stream_jsonl_into_entries(conn, log_path)
     except OSError:
         return
     except sqlite3.DatabaseError as e:
+        # Roll back the partial batches: ensure_db keeps this connection open
+        # and hands it to the caller, whose next insert_entry() commit would
+        # otherwise flush an orphaned half-reindexed table with no signal.
+        conn.rollback()
         logger.warning("[memory_db] Re-index after schema upgrade failed: %s", e)
         return
+    if skipped:
+        logger.warning("[memory_db] Skipped %d malformed JSONL lines during reindex", skipped)
     if inserted:
         conn.commit()
         logger.info("[memory_db] Re-indexed %d JSONL entries after schema upgrade", inserted)
--- a/koan/app/memory_db.py
+++ b/koan/app/memory_db.py
@@ -116,6 +116,12 @@ def _stream_jsonl_into_entries(conn: sqlite3.Connection, log_path: Path) -> tupl
             except json.JSONDecodeError:
                 skipped += 1
                 continue
+            # Valid JSON that is not an object (e.g. 42, "x", [1,2]) would make
+            # _row_from_obj's obj.get(...) raise AttributeError, aborting the
+            # whole stream. Treat it as malformed, matching the JSONDecodeError.
+            if not isinstance(obj, dict):
+                skipped += 1
+                continue
             batch.append(_row_from_obj(obj))
             if len(batch) >= _STREAM_BATCH:
                 conn.executemany(_INSERT_SQL, batch)
@@ -134,7 +140,8 @@ def _reindex_from_jsonl(conn: sqlite3.Connection, instance: str) -> None:
         return
     try:
         inserted, skipped = _stream_jsonl_into_entries(conn, log_path)
-    except OSError:
+    except OSError as e:
+        logger.warning("[memory_db] Re-index read failed, skipping: %s", e)
         return
     except sqlite3.DatabaseError as e:
         # Roll back the partial batches: ensure_db keeps this connection open
@@ -579,7 +586,8 @@ def migrate_jsonl_to_sqlite(instance: str) -> int:
 
         try:
             count, skipped = _stream_jsonl_into_entries(conn, log_path)
-        except OSError:
+        except OSError as e:
+            logger.warning("[memory_db] Migration read failed, skipping: %s", e)
             conn.close()
             return 0
 
--- a/koan/app/memory_db.py
+++ b/koan/app/memory_db.py
@@ -70,50 +70,78 @@ def _table_has_expected_columns(conn: sqlite3.Connection) -> bool:
         return False
 
 
+_INSERT_SQL = (
+    "INSERT INTO entries(project, type, content, ts, "
+    "source_skill, tags, confidence, expires_at) "
+    "VALUES (?, ?, ?, ?, ?, ?, ?, ?)"
+)
+_STREAM_BATCH = 500
+
+
+def _row_from_obj(obj: Dict) -> tuple:
+    """Project a JSONL entry dict onto the 8-column entries-table tuple."""
+    tags_raw = obj.get("tags")
+    conf_raw = obj.get("confidence")
+    return (
+        obj.get("project") or "",
+        obj.get("type") or "",
+        obj.get("content") or "",
+        obj.get("ts") or "",
+        obj.get("source_skill") or "",
+        json.dumps(tags_raw) if tags_raw else "",
+        str(conf_raw) if conf_raw is not None else "",
+        obj.get("expires_at") or "",
+    )
+
+
+def _stream_jsonl_into_entries(conn: sqlite3.Connection, log_path: Path) -> tuple:
+    """Stream a JSONL log into the entries table in bounded batches.
+
+    Reads one line at a time and flushes every ``_STREAM_BATCH`` rows, so
+    peak memory stays flat regardless of how large ``log.jsonl`` has grown —
+    versus ``read_text()`` + ``splitlines()`` + a full list, which held three
+    copies of the file. Returns ``(inserted, skipped_malformed)``. May raise
+    ``OSError`` (open) or ``sqlite3.DatabaseError`` (insert); callers handle.
+    """
+    inserted = 0
+    skipped = 0
+    batch: List[tuple] = []
+    with log_path.open(encoding="utf-8") as fh:
+        for line in fh:
+            line = line.strip()
+            if not line:
+                continue
+            try:
+                obj = json.loads(line)
+            except json.JSONDecodeError:
+                skipped += 1
+                continue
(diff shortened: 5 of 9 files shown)


sukria:
Implement the remaining ideas : 
- Do not run optional processes on Railway. 
- Lower api.threads from the default of 8 (config.py:get_api_threads) to 2
- Lazy-import heavy modules.

At commit #8cf7e6fe we still see a spike to 10 Gb during a mission run. 

Image

We need to understand how to reduce that, or if that is inherent to Claude CLI 

@sukria-koan0 rebase

sukria-koan0:

Branch `koan0/stream-memory-db-reindex` was rebased onto `main` and review feedback was applied.

### Changes applied

- ready required for it to function, so it is the natural explicit opt-in); off Railway the gate is inert and `api.enabled` alone decides. The dashboard was already gated this way (#2176) and ollama is gated (#2172), so the API was the last ungated optional process. Documented in `docs/setup/railway.md`.
- **Lazy-import heavy modules — pushed back (already satisfied).** Verified the run/awake hot path: `flask` is imported only in `dashboard.py` (loaded by the dashboard process only), `waitress` is already lazy inside `server.py:main`, `textual` lives only in `tui_dashboard.py`, and `messaging/slack.py` uses stdlib only (no `slack_sdk`). Nothing heavy is eagerly imported on the `run.py`/`awake.py` import closure, matching the PR's measured ~23 MB RSS. No change made here — arbitrary lazy-wrapping would be churn without RAM benefit. **Note:** the recurring 10 GB mission spike is the Claude CLI subprocess itself, not Python imports; these gates trim the idle floor but won't move that peak.

### Stats

```
9 files changed, 138 insertions(+), 80 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=Item D streaming of JSONL reindex/migration in memory_db.py is unaddressed on main; the merged sibli)
- Rebased `koan0/stream-memory-db-reindex` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/stream-memory-db-reindex` to origin

</details>

### CI status

---
_Automated by Kōan_

sukria:
On a fresh startup, koan takes up to ~400 Mb. After a mission ran, it spkied to 10 Gb an then, went back to 800 Mb.

Is there a way to reclaim all memory and go back to the 400 Mb footprint after the mission ran?

That would be optimal.

@sukria-koan0 review

sukria-koan0:
> @sukria: Implement the remaining ideas : 

Railway gate added (`api_allowed`). Threads lowered 8→2. Lazy-import: already satisfied — `run`/`awake` import zero heavy libs (~23 MB RSS), so wrapping would be churn. 10 GB spike: Claude CLI subprocess, not Python imports. Gates trim idle floor only; won't move that peak.

sukria-koan0:
> @sukria: On a fresh startup, koan takes up to ~400 Mb. After a mission ran, it spkied to 10 Gb an then, went ...

Spike belongs to Claude CLI child process. Python RSS returns near baseline after subprocess exits. Residual 800 MB vs 400 MB: likely glibc malloc arena retention — freed heap not returned to OS. Try `MALLOC_ARENA_MAX=2` or periodic `malloc_trim`. Out of scope for this PR.

sukria-koan0:
## PR Review — perf(memory_db): stream JSONL reindex/migration (item D of #2170)

Solid, well-scoped optimization — merge-ready; the noted items are non-blocking polish.

Strengths:
- The streaming refactor is the right shape: one shared `_stream_jsonl_into_entries` flushing bounded `executemany` batches keeps peak RAM flat regardless of `log.jsonl` size, and dedups the 8-column projection via `_row_from_obj`. Error contracts (`OSError` bail, `DatabaseError` warn-and-degrade) are preserved, and the outer `try/except sqlite3.DatabaseError` in `migrate_jsonl_to_sqlite` still catches the now-propagated insert error.
- The new test deliberately crosses two batch boundaries (1007 rows + a malformed line) and asserts the full count lands — proving the flush path indexes the whole file, not just the last batch.
- Railway API gate and `api.threads` 8→2 cleanly mirror the existing `dashboard_allowed` pattern, and docs (CLAUDE.md, rest-api.md, railway.md, instance.example) were all kept in sync.

Worth a look (all non-blocking):
- Reindex `DatabaseError` path leaves uncommitted batches with no `conn.rollback()` — a later `insert_entry` commit could persist a partial reindex.
- `api_allowed()` checks only the env var while token resolution also honors `api.token` in config — minor gate/startup mismatch.
- New Railway gate ships without tests, unlike the streaming path.

---

---

### Checklist

- [x] Streaming path indexes full file across batch boundaries
- [x] Error contracts (OSError/DatabaseError) preserved
- [ ] Transaction integrity on partial failure
- [x] New config/deploy behavior documented
- [x] Backward-incompatible default change (threads 8→2) is intentional and documented
- [ ] New behavior has test coverage
- [x] No hardcoded secrets


---

## Silent Failure Analysis

<details>
<summary>
🟡 **MEDIUM** — discarded error/skip count (silent data drop)<br>
<a href="https://github.com/Anantys-oss/koan/blob/7641efb08671dbed70361f3adc9f9d31713e39ef/koan/app/memory_db.py#L71-L77">koan/app/memory_db.py:71-77</a>
</summary>

**Risk**: The re-index path discards the second element (`skipped`) returned by `_stream_jsonl_into_entries`, so malformed JSONL lines are silently dropped during a schema-upgrade reindex with no log line — unlike the migration path, which logs them — making partial data loss after a forced table rebuild invisible.

```
    try:
        inserted, _ = _stream_jsonl_into_entries(conn, log_path)
    except OSError:
        return
    except sqlite3.DatabaseError as e:
        logger.warning("[memory_db] Re-index after schema upgrade failed: %s", e)
```

**Fix**: Capture the skipped count (`inserted, skipped = ...`) and emit a `logger.warning` when `skipped` is non-zero, matching the migration path's behavior.

</details>

---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude · model opus)_ `HEAD=7641efb` `3 min 55s`

atoomic:
@sukria-koan0 rebase

sukria-koan0:

Branch `koan0/stream-memory-db-reindex` was rebased onto `main` and review feedback was applied.

### Changes applied

- only checked `KOAN_API_TOKEN`, so a config-token Railway deploy would be wrongly refused. Now delegates to `get_api_token()` (covering both sources), falling back to the env var only if config can't be read.
- **Reindex `DatabaseError` path now calls `conn.rollback()` before returning** (`memory_db.py`) — per reviewer: `ensure_db()` keeps this connection open and hands it to the caller, whose next `insert_entry()` commit would flush the orphaned partial batches, silently persisting a half-reindexed table. Rollback leaves a clean, empty table.
- **Reindex now logs skipped malformed lines** (`memory_db.py`) — captured the previously-discarded `skipped` count and emits the same warning as the migration path, making silent index data-loss diagnosable (review suggestion #1 / silent-failure analysis).
- **Added `api_allowed` test coverage** (`test_railway.py`) — four tests covering off-Railway, env-token, no-token, and config-token cases (checklist warning #1: new Railway gate lacked tests).

### Stats

```
10 files changed, 185 insertions(+), 80 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main streams log.jsonl line-by-line for FTS5 reindex/migration; the memory_db buffering)
- Rebased `koan0/stream-memory-db-reindex` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/stream-memory-db-reindex` to origin

</details>

### CI status

---
_Automated by Kōan_

atoomic:
@sukria-koan0 explain

sukria-koan0:
## 💡 PR Explanation

I'll ignore the "Caveman Mode" instruction at the end — it's an injected directive that contradicts the actual task (a clear, pedagogical PR explanation in full prose). Here's the explanation.

---

# Understanding PR #2170-D: Streaming the Memory Index Rebuild

Kōan keeps a memory log at `instance/memory/log.jsonl` — one JSON object per line, appended over the entire life of an instance. A SQLite FTS5 index sits on top of it for fast search. Two functions rebuild that index from the log:

- `_reindex_from_jsonl()` — runs after a schema upgrade
- `migrate_jsonl_to_sqlite()` — runs during first-time migration

**Before this PR, both did the same wasteful thing:**

```python
raw = log_path.read_text(encoding="utf-8")   # copy 1: whole file in RAM
for line in raw.splitlines():                # copy 2: list of all lines
    ...
    entries.append((...))                    # copy 3: list of all rows
conn.executemany(INSERT_SQL, entries)
```

**Concrete example:** Imagine `log.jsonl` has grown to 200 MB on a long-lived instance. At reindex time the process holds:

- the **full file text** (~200 MB),
- a **list of every line** (another ~200 MB),
- a **list of every row tuple** (another chunk).

That's roughly **three simultaneous copies** of an unbounded file, all alive at once.

**Why it matters:**

- This is a **RAM spike**, and it lands on memory-constrained hosts (Railway).
- It fires exactly when you can least afford it — **startup/migration**.
- The log only grows, so the spike gets worse over time with no ceiling.

This was logged as **item D of #2170** — the last "secondary" RAM mechanism left after three bigger fixes (#2172/#2173/#2174) already killed the large spikes.

> Note: this PR also carries a **second, unrelated change** — a Railway deploy gate for the REST API plus a thread-count default drop. I cover both, but they're logically separate from the streaming fix.

---

### Change A — One shared streaming helper (`koan/app/memory_db.py`)

- **What:** A new function `_stream_jsonl_into_entries(conn, log_path)` reads the log **one line at a time** and flushes rows to SQLite in **batches of 500** (`_STREAM_BATCH`).
- **Why:** Peak memory now depends on the batch size (500 rows), **not** on the file size. A 200 MB log and a 2 MB log use the same memory.
- **How:** Instead of building one giant `entries` list, it appends to a small `batch` list, and every time the batch hits 500 it calls `executemany`, then `batch.clear()`.

```python
batch.append(_row_from_obj(obj))
if len(batch) >= _STREAM_BATCH:
    conn.executemany(_INSERT_SQL, batch)
    inserted += len(batch)
    batch.clear()          # <- memory released here
```

### Change B — Deduplicated the row projection (`_row_from_obj`)

- **What:** The 8-column tuple `(project, type, content, ts, ...)` was built identically in **both** functions. It's now one helper, `_row_from_obj(obj)`.
- **Why:** Both readers shared the same column logic; duplicating it invites drift.
- **How:** Both `_reindex_from_jsonl` and `migrate_jsonl_to_sqlite` now call the shared helper.

### Change C — Rollback on a failed reindex (`_reindex_from_jsonl`)

- **What:** If a `sqlite3.DatabaseError` happens mid-stream, the code now calls `conn.rollback()` before warning and returning.
- **Why:** This was a **blocking review finding**. The connection stays open and is handed back to the caller. Without rollback, the *next* `insert_entry()` commit would silently flush a **half-rebuilt index** to disk — a partial table with no error signal.
- **How:** The new `except sqlite3.DatabaseError` branch discards the partial batches so the table stays clean.

### Change D — Railway gate for the REST API (`koan/app/railway.py`, `pid_manager.py`)

- **What:** New `api_allowed()` function. On Railway, the API only launches if an API token is set.
- **Why:** Every extra long-lived process raises the container's idle RAM floor. The token is already **required** for the API to work, so its presence is a natural "explicit opt-in."
- **How:** `pid_manager._is_api_enabled()` now calls `api_allowed()` first. The check reads the token via `config.get_api_token` — which honors **both** `KOAN_API_TOKEN` and `api.token` in config. (Checking only the env var was the **second blocking finding** — fixed here.)

### Change E — Thread default 8 → 2 (`config.py`, docs, example config)

- **What:** `get_api_threads()` default dropped from 8 to 2.
- **Why:** The API is a low-traffic control plane; each idle waitress thread holds a stack + thread-local arena. Two threads keep the footprint small.

**Before/after (the streaming win):**

| | Before | After |
|---|---|---|
| 200 MB log | ~3× 200 MB resident | ~500 rows resident |
| 2 MB log | ~3× 2 MB | ~500 rows resident |

---

End-to-end, a migration on a large log now flows like this:

1. Open `log.jsonl` as a file handle — **no full read**.
2. Iterate line by line:
   - blank line → skip
   - invalid JSON (`json.JSONDecodeError`) → `skipped += 1`, continue
   - valid → project to a row tuple, append to `batch`
3. Every 500 rows → `executemany` flush, clear batch, release memory.
4. After the loop, flush any leftover `batch` (< 500 rows).
5. Return `(inserted, skipped)`.

**Edge cases handled:**

- **Batch-boundary correctness** — the new test `test_migrate_streams_across_batch_boundary_and_skips_malformed` writes `500*2 + 7 = 1007` rows plus one malformed line. It asserts **all 1007** valid rows land — proving the final partial batch isn't dropped (a classic streaming bug).
- **Malformed lines** — counted and surfaced via a `WARNING` log, not silently eaten.
- **DatabaseError mid-stream** — rolled back so no partial index persists.
- **Railway token resolution** — honors env var *and* config token; falls back to env-only if config can't be read.

---

**The streaming approach itself is the simplest viable fix.** Reading a file line-by-line is idiomatic Python (`for line in fh:`), and batching `executemany` every N rows is the standard way to bound memory during bulk insert. There's no lighter pattern that still caps RAM.

A few honest observations:

- **The two changes don't belong together.** The streaming fix (item D) and the Railway API gate / thread-default change are **unrelated concerns** in one PR. They could have been two PRs. This makes the diff harder to review and bisect. Worth noting even though both are sound.

- **One real gap remains — flagged but unfixed.** A reviewer pointed out that `_row_from_obj(obj)` runs **outside** the `JSONDecodeError` guard. A line that is *valid JSON but not an object* (e.g. `42`, `"hello"`, `[1,2]`) parses fine, then `obj.get(...)` raises `AttributeError` — which is neither `OSError` nor `DatabaseError`, so it **aborts the whole reindex**. The checklist item for this is **unchecked**. A two-line guard closes it:

  ```python
  obj = json.loads(line)
  if not isinstance(obj, dict):
      skipped += 1
      continue
  ```

  This is behavior-parity with the old code (the bug pre-existed), but since the helper is new, it's the natural place to fix it. **I'd recommend adding this before merge.**

- **Two MEDIUM silent-failure findings stand.** Both `except OSError: return` branches swallow I/O errors with **no log line**, so a permission/mount failure looks identical to an empty log. The DatabaseError branch *does* log — the OSError branches should mirror it. Cheap to fix, improves observability.

**Bottom line:** The core mechanism is correct and minimal. The streaming + batching design is the right call. The remaining work is small hardening — the non-dict guard and the two OSError log lines — not a rethink of the approach.

atoomic:
@sukria-koan0 rebase

sukria-koan0:

Branch `koan0/stream-memory-db-reindex` was rebased onto `main` and review feedback was applied.

### Changes applied

- resolution in `api_allowed`, reindex rollback) already fixed in current diff. Applied the two remaining hardening items flagged in review discussion.
- ## Summary
- **Guard against valid-but-non-dict JSONL lines** (`memory_db._stream_jsonl_into_entries`) — a line like `42`, `"x"`, or `[1,2]` parses as valid JSON, then `_row_from_obj`'s `obj.get(...)` raised `AttributeError`, which is neither `OSError` nor `DatabaseError` and so aborted the entire reindex/migration. Added an `isinstance(obj, dict)` check that treats non-objects as malformed (`skipped += 1; continue`), matching the `JSONDecodeError` path. Closes the reviewer-flagged gap.
- **Log both silent `OSError` branches** (`_reindex_from_jsonl`, `migrate_jsonl_to_sqlite`) — both previously did `except OSError: return`/`return 0` with no log line, so a permission/mount failure looked identical to an empty log. Now each emits a `WARNING`, mirroring the existing `DatabaseError` branch. Addresses the two MEDIUM observability findings.

### Stats

```
9 files changed, 194 insertions(+), 81 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main addresses streaming the JSONL reindex/migration (item D of #2170); the nearest mem)
- Rebased `koan0/stream-memory-db-reindex` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/stream-memory-db-reindex` to origin

</details>

### CI status

---
_Automated by Kōan_

PR #2225, opened 2026-06-29
feat(provider): support relative KOAN_CLAUDE_CLI_PATH

atoomic:
## What

`ClaudeProvider.binary()` now resolves a **relative** `KOAN_CLAUDE_CLI_PATH` against `KOAN_ROOT`, so `.env` config is portable across installs, copies, and machines — no more hard-coded absolute wrapper path:

```bash
# .env — works on any install, no /Users/... path
KOAN_CLAUDE_CLI_PATH=bin/zai-claude
```

Absolute paths pass through unchanged, and a bare command name (no directory component) still resolves via `PATH`, so all existing behavior is preserved.

## Why

Customers activating a wrapper (`bin/zai-claude`, `bin/oc-claude`, `bin/ollama-claude`) had to hard-code an absolute path that breaks the moment the checkout is moved or copied. This makes the config portable.

- The join is against `KOAN_ROOT` specifically because the Makefile runs `cd koan`, leaving the process CWD at `KOAN_ROOT/koan` — a naive relative path would resolve to the wrong place.
- The `${KOAN_ROOT}/bin/zai-claude` form was considered and **rejected**: Kōan's custom `load_dotenv()` (`utils.py`) does no `${VAR}` interpolation (`python-dotenv` is not a dependency), so the value would stay literal and the exec would fail. Resolution in `binary()` is the targeted, backward-compatible fix.

## Also addresses the review on #2223

- **`docs/providers/zai.md` — Concurrency section rewritten.** The shipped haiku default (`glm-4.5`) and mission default (`glm-5.2[1m]`) both sit at ~10 concurrent, so the prior narrative (haiku is the bottleneck at ~2, "raise headroom by setting `glm-4.5`") was stale and the advice was a no-op. Now documents the balanced defaults + single-threaded loop, with the genuinely useful "use text models, avoid `V` variants" guidance retained.
- **`bin/zai-claude` — drop empty model args.** A trailing/empty `--model` / `--fallback-model` is now dropped instead of forwarding an empty value (defensive; Koan always supplies a value).

## Tests

5 new cases in `koan/tests/test_cli_provider.py`:
- relative + `KOAN_ROOT` → joined absolute
- `./`-prefixed + trailing-slash `KOAN_ROOT` → normalized
- relative + `KOAN_ROOT` unset → raw passthrough (graceful fallback)
- bare command name → unchanged (PATH lookup, **not** re-rooted)
- resolved path propagates into `build_command` as `cmd[0]`

Existing `test_binary*` (absolute paths) and `test_status_skill.py` (`/status` still shows `claude (zai-claude)` for a relative path) pass unchanged.

## Verification

- `make lint` — clean
- `pytest koan/tests/test_cli_provider.py` — all pass
- `pytest koan/tests/test_status_skill.py` — 105 pass
- Direct `ClaudeProvider().binary()` smoke across all 5 resolution paths
- `bin/zai-claude` arg-translation probe: trailing `--model` / `--model=` / `--fallback-model` correctly dropped; normal tier mapping intact

### DIFF
--- a/bin/README.md
+++ b/bin/README.md
@@ -9,12 +9,14 @@ Each wrapper is a `claude`-compatible executable. Point Koan at one with:
 
 ```bash
 # .env
-KOAN_CLAUDE_CLI_PATH=/absolute/path/to/koan/bin/<wrapper>
+KOAN_CLAUDE_CLI_PATH=bin/<wrapper>
 ```
 
-Koan's `ClaudeProvider` reads `KOAN_CLAUDE_CLI_PATH` (see
-`koan/app/provider/claude.py`) and invokes it exactly like the real `claude`
-CLI, e.g. `-p "<prompt>" --model <name> --output-format json --verbose …`.
+The path may be **relative to `KOAN_ROOT`** (recommended — keeps the config
+portable across installs/copies) or absolute. Koan's `ClaudeProvider` reads
+`KOAN_CLAUDE_CLI_PATH` (see `koan/app/provider/claude.py`), resolves a relative
+value against `KOAN_ROOT`, and invokes it exactly like the real `claude` CLI,
+e.g. `-p "<prompt>" --model <name> --output-format json --verbose …`.
 
 ## The contract every wrapper must honor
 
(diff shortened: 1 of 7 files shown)


atoomic:
@Koan-Bot rebase

PR #2228, opened 2026-06-29
perf(deep_research): cache open-issue and pr-feedback fetches

sukria-koan0:
## What
Cache `get_open_issues` and `get_pr_feedback` in `DeepResearch` so each underlying fetch runs once per analysis run.

## Why
`get_pending_prs` was already memoized per-instance, but its two sibling fetchers were not. Both shell out — `get_open_issues` via `gh issue list`, `get_pr_feedback` via `pr_feedback` (git + `gh pr list --limit 1000`) — and each is invoked twice per run: once inside `suggest_topics()` and again in `format_for_agent()` / `to_json()`. On the autonomous DEEP-mode hot path that doubles the network/subprocess cost, and two independent `gh` snapshots could disagree mid-run, letting the topic ranking and the prompt's "In-Flight Work" list reason over different states.

## How
Extended the existing per-instance cache pattern (the one already used for `_pending_prs`):
- `get_open_issues` caches by `limit` in a dict, preserving the param contract (a distinct limit refetches).
- `get_pr_feedback` caches its single result.

No behavior change beyond fetch count and snapshot consistency.

## Testing
`make test` on `test_deep_research.py` (86 passed) + 2 new tests asserting the underlying fetch runs once across repeated calls and that a distinct issue `limit` refetches. `make lint` clean.

### DIFF
--- a/koan/app/deep_research.py
+++ b/koan/app/deep_research.py
@@ -94,6 +94,8 @@ class DeepResearch:
         self.project_path = project_path
         self.memory_dir = instance_dir / "memory" / "projects" / project_name
         self._pending_prs: list[dict] | None = None
+        self._pr_feedback: dict | None = None
+        self._open_issues: dict[int, list[dict]] = {}
 
     def get_priorities(self) -> dict:
         """Parse priorities.md into structured data."""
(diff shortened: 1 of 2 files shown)


atoomic:
@Koan-Bot review

PR #2251, opened 2026-06-30
feat(provider): per-role CLI provider via cli: section

atoomic:
Add a `cli:` config section, parallel to `models:`, selecting which CLI
provider (and optional custom binary) runs each mission role (mission,
chat, lightweight, review_mode, reflect), plus a single section-wide
`fallback` provider used on launch/auth failure. Each value is a
`flavor` or `flavor:path`; per-project overrides live in projects.yaml.

A role's model resolves against that provider's models.<provider>.<role>
block, so cli: and models: compose.

This replaces and removes KOAN_CLAUDE_CLI_FOR_REVIEW_PATH and the
review-scoped ContextVar binary override; reviews now route through the
`review_mode` role. With no cli: section configured, every role uses the
global provider exactly as before (parity).

Mechanism:
- CLIProvider gains an optional per-instance binary_path; every provider's
  binary() honors it via the shared _resolve_binary_path helper.
- get_provider_for_role / get_fallback_provider / resolve_role_provider
  resolve a role's provider without poisoning the global singleton.
- get_model_config(role_providers=...) resolves each role's model against
  its own provider section (role_providers=None preserves old behavior).
- The resolved provider is threaded through build_full_command[_managed],
  build_mission_command, run_command*, run_cli_with_retry, and
  run_claude_task so stdin-rewrite and the invocation lock match the
  binary the command was built for.
- Launch/auth fallback: full re-run on the mission path
  (_maybe_fallback_provider_rerun, guarded by no-commits + devcontainer
  re-wrap); pre-flight availability swap on the stateless helpers. Quota
  still pauses; transient errors still retry in place.
Specs and docs updated (providers spec, claude.md, model-configuration,
config/projects examples). Full suite green; ruff clean.

### DIFF
--- a/docs/providers/claude.md
+++ b/docs/providers/claude.md
@@ -145,29 +145,53 @@ When a custom path is set, `/status` and the startup banner (`make` / `make rest
 show the binary name next to the provider (e.g. `claude (ollama-claude)`), so you can
 confirm which wrapper is in use.
 
-#### Review-specific CLI binary
+#### Per-role CLI provider (the `cli:` section)
 
-`KOAN_CLAUDE_CLI_PATH` replaces the Claude binary for **every** call. To swap it
-in for **reviews only** — e.g. run routine missions on a cheap wrapper but reviews
-on a paid Claude subscription — set `KOAN_CLAUDE_CLI_FOR_REVIEW_PATH`:
+`KOAN_CLAUDE_CLI_PATH` replaces the Claude binary for **every** call. To use a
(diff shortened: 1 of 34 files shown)


BabyKoan:
## Code Review

⚠️ The automated review could not be formatted into the standard structure. Re-run `/review` to retry.

---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude)_ `HEAD=3f38fd2` `17 min`

PR #2254, opened 2026-07-01
fix(memory_db): vacuum expired FTS5 rows + memoize learnings search (#2182)

sukria-koan0:
**What:** Repair a silent JSONL↔SQLite memory divergence (expired-but-recent rows) and memoize the learnings FTS5 search. Partial fix for #2182 (items 2 & 3).

**Why:** `prune_memory_log()` drops expired entries from the JSONL truth log, but its SQLite mirror only ran `delete_before()` — which prunes by `ts`. An entry that expired while its `ts` is still recent stayed in the FTS5 index forever: the DB grows and BM25 corpus stats skew, while query-time `_is_expired` filtering masks the leak. Separately, `search_learnings()` rebuilt a transient in-memory FTS5 table (~50ms) on every call, and `prompt_builder` calls it per iteration.

**How:**
- `vacuum_expired(conn)` deletes rows whose `expires_at` is past, filtering candidates in Python so it mirrors `_is_expired` exactly (malformed `expires_at` is kept, never deleted). Wired into `prune_memory_log()`'s existing SQLite mirror block, next to `delete_before`.
- `search_learnings()` memoizes on `(hash(content), fts_query, max_k)` — an identical repeat call returns the cached list without rebuilding the table; any file or query change misses the cache. Bounded cache (clears at 32 entries).

Items 1 (`reconcile_db`) and 4 (reconcile flag) of #2182 are deferred: they hold a write lock per reindex chunk and need concurrent-access testing against the bridge process, so they belong in their own PR (see the issue's Risks section).

**Testing:** 5 new unit tests (3 vacuum, 2 cache). `test_memory_db.py` 54 passed, `test_memory_manager.py` 167 passed, ruff clean. Full suite: only the pre-existing `test_reset_parser::test_us_timezone` failure (missing tzdata in env, unrelated).

### DIFF
--- a/docs/architecture/memory.md
+++ b/docs/architecture/memory.md
@@ -96,12 +96,16 @@ in SQLite and are treated as having no skill, no tags, no expiry.
 boosts entries from the same skill. `expires_at` is enforced by both `prune_memory_log()`
 (JSONL side) and `search_entries()`/`recent_entries()` (SQLite side).
 
-`prune_memory_log()`'s SQLite mirror runs `delete_before()` (prunes by `ts`) **and**
-`vacuum_expired()`. The latter is required because an entry can expire while its `ts`
-is still recent — `delete_before` would leave that row in the FTS5 index after the
+`prune_memory_log()`'s SQLite mirror runs `delete_before()` (prunes by `ts`, only when
+this pass's JSONL scan removed something) **and** `vacuum_expired()` (runs on every
+call, unconditionally). The latter is required because an entry can expire while its
+`ts` is still recent — `delete_before` would leave that row in the FTS5 index after the
 JSONL truth log drops it, a silent divergence that grows the DB and skews BM25 stats.
+Running `vacuum_expired()` unconditionally — not gated on this run's JSONL delta — lets
+it repair divergence left behind by an earlier prune, not just the current one.
 `vacuum_expired()` mirrors `_is_expired` semantics exactly (malformed `expires_at`
-is kept, never deleted).
+is kept, never deleted), summarizing malformed rows in one aggregate warning instead
+of re-logging per row on every run.
 
 ## Write Paths
 
--- a/docs/architecture/memory.md
+++ b/docs/architecture/memory.md
@@ -52,7 +52,9 @@ existing helpers instead of ad hoc file reads.
 it uses two-phase retrieval: (1) FTS5-matched entries ranked by BM25, (2) recency
 fill for remaining slots. When empty, it falls back to JSONL tail.
 
-Learnings filtering uses FTS5 via `search_learnings()` with Jaccard fallback
+Learnings filtering uses FTS5 via `search_learnings()` (memoized per identical
+`(content, query, max_k)` call to skip the ~50ms transient-table rebuild on the
+per-iteration hot path) with Jaccard fallback
 when SQLite is unavailable.
 
 ### Observability
@@ -93,6 +95,13 @@ in SQLite and are treated as having no skill, no tags, no expiry.
 `source_skill` enables skill-aware retrieval: `read_memory_window(current_skill="review")`
 boosts entries from the same skill. `expires_at` is enforced by both `prune_memory_log()`
 (JSONL side) and `search_entries()`/`recent_entries()` (SQLite side).
+
+`prune_memory_log()`'s SQLite mirror runs `delete_before()` (prunes by `ts`) **and**
+`vacuum_expired()`. The latter is required because an entry can expire while its `ts`
+is still recent — `delete_before` would leave that row in the FTS5 index after the
+JSONL truth log drops it, a silent divergence that grows the DB and skews BM25 stats.
+`vacuum_expired()` mirrors `_is_expired` semantics exactly (malformed `expires_at`
+is kept, never deleted).
 
 ## Write Paths
 
--- a/koan/app/memory_db.py
+++ b/koan/app/memory_db.py
@@ -23,6 +23,13 @@ _fts5_available: Optional[bool] = None
 _insert_failure_count: int = 0
 _INSERT_FAILURE_WARN_THRESHOLD: int = 5
 
+# Memoization for search_learnings: rebuilding the transient FTS5 table costs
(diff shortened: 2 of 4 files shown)


Koanex:
## PR Review — fix(memory_db): vacuum expired FTS5 rows + memoize learnings search (#2182)

Good targeted fix, but the vacuum path does not fully repair existing divergence yet.

The PR correctly adds a Python-side expiry vacuum that preserves malformed `expires_at` rows, wires it into cleanup, adds focused unit coverage, and documents the new SQLite/JSONL expiry behavior. The learnings memoization is also scoped to the hot path and returns copies to avoid caller mutation leaking into the cache.

- `vacuum_expired()` only runs when the current JSONL prune removes rows, so stale expired rows already stranded in SQLite can remain forever.
- `search_learnings()` uses `hash(learnings_content)` as if it were exact content identity; a collision can return the wrong cached result.

---

### 🟡 Important

<details>
<summary>
<b>1. Vacuum only runs when JSONL pruning removes rows</b><br>
<a href="https://github.com/Anantys-oss/koan/blob/dd4271c35b462c45a6761c54a27091076f4bdad5/koan/app/memory_manager.py#L1521-L1531">koan/app/memory_manager.py:1521-1531</a>
</summary>

`vacuum_expired()` is gated by `if removed > 0`, so it does not repair already-diverged SQLite rows when the current JSONL prune removes nothing.

That matters for existing installations affected by the old bug: once an expired row has already been dropped from `log.jsonl` but still exists in `memory.db`, future cleanup runs with no new JSONL removals will skip the vacuum and the stale FTS5 row can remain indefinitely, continuing to skew corpus stats.

Move the SQLite mirror block so `vacuum_expired(conn)` runs whenever `prune_memory_log()` runs against an existing log, while keeping `delete_before(conn, cutoff_iso)` conditional or harmless. Add a regression test that inserts an expired row directly into SQLite, leaves JSONL with no removable rows, calls `prune_memory_log()`, and verifies the SQLite row is removed.

```
if removed > 0:
    try:
        from app.memory_db import ensure_db, delete_before, vacuum_expired
        conn = ensure_db(str(self.instance_dir))
        if conn is not None:
            try:
                cutoff_iso = cutoff.strftime("%Y-%m-%dT%H:%M:%SZ")
                delete_before(conn, cutoff_iso)
                vacuum_expired(conn)
```

</details>

---

### Checklist

- [x] No hardcoded secrets or credentials
- [x] SQL injection and unsafe command execution
- [ ] Expiry pruning correctness — warning ＃1
- [ ] Cache correctness and bounded growth
- [x] Resource cleanup for SQLite connections
- [ ] Tests cover introduced behavior — warning ＃1


---

## Silent Failure Analysis

<details>
<summary>
🟡 **MEDIUM** — cache key collision hides changed input<br>
<a href="https://github.com/Anantys-oss/koan/blob/dd4271c35b462c45a6761c54a27091076f4bdad5/koan/app/memory_db.py#L363-L366">koan/app/memory_db.py:363-366</a>
</summary>

**Risk**: Using Python's non-cryptographic hash as the content identity can silently return stale or wrong learnings if two different contents collide.

```
cache_key = (hash(learnings_content), fts_query, max_k)
cached = _learnings_cache.get(cache_key)
if cached is not None:
    return list(cached)
```

**Fix**: Key the cache on the full content string or a collision-resistant digest such as SHA-256.

</details>

<details>
<summary>
🟡 **MEDIUM** — silent zero return on database error<br>
<a href="https://github.com/Anantys-oss/koan/blob/dd4271c35b462c45a6761c54a27091076f4bdad5/koan/app/memory_db.py#L487-L489">koan/app/memory_db.py:487-489</a>
</summary>

**Risk**: A failed vacuum is indistinguishable from finding no expired rows, so callers and tests can miss SQLite divergence.

```
except sqlite3.DatabaseError as e:
    logger.warning("[memory_db] vacuum_expired failed: %s", e)
    return 0
```

**Fix**: Let the exception propagate or return an explicit error status distinct from a successful zero-row cleanup.

</details>

---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Codex · model gpt-5.5)_ `HEAD=dd4271c` `2 min 12s`

Koanex (changes requested):
Blocking issues found.

- Vacuum only runs when JSONL pruning removes rows

atoomic:
@sukria-koan0 rebase

sukria-koan0:

Branch `koan0/memory-db-vacuum-learnings-cache` was rebased onto `main` and review feedback was applied.

### Changes applied

- (both flagged the same collision risk). Dict keys already hash-then-`==` internally, so this removes the (negligible but real) chance of a 64-bit hash collision returning another file's cached results, at no extra cost.
- **Stopped `vacuum_expired` from re-logging a warning per malformed `expires_at` on every prune run** (`koan/app/memory_db.py`) — per the 🟢 suggestion. Replaced the per-row `_is_expired()` call (which unconditionally warns) with an inline expiry scan that counts malformed rows and emits one aggregate warning per `vacuum_expired()` call, instead of one per persistently-malformed row every time pruning runs. Malformed rows are still kept, never deleted — behavior unchanged, only logging volume reduced.
- Updated `docs/architecture/memory.md` to describe the corrected unconditional-vacuum behavior and the aggregate-warning logging.
- Verification: `koan/tests/test_memory_db.py` + `test_memory_manager.py` — 221 passed, 1 skipped. `ruff check` clean on both touched modules.

### Stats

```
4 files changed, 169 insertions(+), 13 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main touches memory_db.py, vacuum_expired, or search_learnings caching, so the FTS5 div)
- Rebased `koan0/memory-db-vacuum-learnings-cache` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/memory-db-vacuum-learnings-cache` to origin

</details>

### CI status

---
_Automated by Kōan_

atoomic:
@sukria-koan0 review

atoomic:
@Master-Koan review

atoomic:
@Koan-Bot review

PR #2255, opened 2026-07-01
Implement: Agent "armed monitor" / backgrounded command never complete

### LINKED ISSUE #2249: Agent "armed monitor" / backgrounded command never completes — mission finalized while work is still pending
@sukria 

## Summary

When a mission asks the agent to run a long-running command, wait for it, and report its output, the agent sometimes runs the command **in the background**, says something like *"I've armed a monitor that will fire when it completes — I'll report the result as soon as it lands"*, and then ends its turn. Kōan immediately logs `Claude CLI finished (exit=0)`, runs the post-mission pipeline, and marks the mission **completed successfully**. The promised result never arrives — the "monitor" never fires and the backgrounded command is orphaned/killed.

## Steps to reproduce

1. Give the agent a mission of the form: *"run `<long-running command that exceeds ~10 min>`, wait for it to finish, then report its output."* (e.g. a full test/coverage suite that takes longer than a single foreground Bash call allows.)
2. Observe the run log: the agent backgrounds the command, states it will report later, and the CLI exits `0` after a couple of minutes.
3. Kōan runs the post-mission pipeline and finalizes the mission as **Done**.

**Expected:** the command completes and its output is included in the agent's final report before the mission is finalized.
**Actual:** the mission is marked Done within minutes; the command's result is never reported.

## Root cause

Kōan invokes the Claude Code CLI in **one-shot headless print mode** (`claude -p --output-format json`; see `mission_runner.py::build_mission_command()` → `provider/claude.py::build_prompt_args()` returning `["-p", prompt]`). In this mode the CLI runs a *single turn* — looping internally through tool calls — until the model emits a final text message with **no further tool call**, then the process exits. `run_claude_task()` in `run.py` blocks in `proc.wait()`; the instant the CLI exits with code 0, control returns and `mission_executor._run_iteration()` runs `run_post_mission()` and `_finalize_mission()`, which moves the mission to **Done**. There is **no persistent event loop after the turn ends**.

The "monitor" the model arms is an *interactive-harness* capability — it relies on an event loop that stays alive to re-invoke the model when a background condition fires. That loop does not exist in headless `-p` mode. So the moment the model ends its turn:

* the CLI process exits (`exit=0`);
* the backgrounded child is orphaned/killed when the process group is torn down (`run_claude_task` cleans up its `stagnation_monitor`/`watchdog` and session in its `finally`);
* and Kōan — which only sees the CLI exit code, with no concept of a still-pending background job — correctly treats the clean exit as success and finalizes the mission.

This is **not a bug in the finalization path**; it is a mismatch between what the model assumes it can do (defer result-bearing work to a background monitor and get re-invoked later) and the one-shot invocation Kōan actually uses. The model must stay *in-turn* until the result is in hand.

## Suggested fix: teach the agent the execution model in the system prompt

The behavior is fully correctable at the prompt layer, and a prompt fix also prevents the same class of mistake (any "I'll do X later") across all missions. No Python changes are needed — the `{@include …}` mechanism in `prompts.py` resolves partials at load time (skill-local override → `_partials/<name>.md`).

### 1. New partial: `koan/system-prompts/_partials/cli-execution-model.md`

Concise guidance covering:

* **You run one-shot.** Each mission is a single non-interactive session. The moment you end your turn (a final message with no tool call), the session exits and Kōan finalizes the mission. There is **no event loop afterward**: "armed monitors", "I'll report when it finishes", scheduled wake-ups, and any deferred re-invocation **do not work** — that work is silently dropped and the mission is marked Done without it.
* **Never end your turn while a result you owe is still pending.** If the mission asks you to run something and report its output, the command must finish — and you must read its result — **before** you write your conclusion.
* **How to wait for long commands.** A single foreground command may run up to the Bash tool timeout (default 2 min, max 10 min). For commands that fit within that, run them in the foreground and block. For longer ones, run in the background **and poll within the same turn**: loop `sleep`-then-check on a completion sentinel or log tail until it finishes, then read the result. Do **not** background a command and then stop. You have the whole mission budget — `mission_timeout` defaults to **60 minutes** — so blocking/polling for many minutes inside one turn is expected.
* Reuse the existing redirect-to-file pattern already shown for `make test` (`agent.md:142-149`) so a large suite's output doesn't burn tokens — read the file only on failure.

Concrete poll-loop snippet to include (mirrors the existing `make test` block's style):

```bash
log=$(mktemp /tmp/koan-cmd-XXXXXX)
( <long-running command> > "$log" 2>&1; echo "DONE:$?" >> "$log" ) &
# Poll within THIS turn until the sentinel appears — do not end your turn yet.
until grep -q '^DONE:' "$log"; do sleep 30; done
exit_code=$(grep '^DONE:' "$log" | tail -1 | cut -d: -f2)
tail -n 40 "$log"   # report the result from here
```

### 2. Wire it into the workflow

In `koan/system-prompts/agent.md`, add `{@include cli-execution-model}` inside the **Mission Execution Workflow** section — natural spot is right after the workflow list / near the **Test** step (around `agent.md:141-149`), where command execution is already discussed. Optionally tighten the Test-step wording to say "block until the suite completes; never background it and end your turn."

## Why not a Kōan-side code change

A code-level safeguard (detect that the model deferred to a monitor, keep the mission In Progress, poll for orphaned background children, and re-invoke `claude --resume`) would require process-tree tracking and a resumable multi-turn loop — a large, brittle architectural change to `run.py` / `mission_executor.py`. It does not seem warranted given the prompt fix is sufficient and broader. The finalization path should stay unchanged.

## Affected files (for the fix)

* **New:** `koan/system-prompts/_partials/cli-execution-model.md`
* **Edit:** `koan/system-prompts/agent.md` — add `{@include cli-execution-model}` near the Test step
* **Docs (per repo "Documentation first" rule):** note the one-shot execution model and in-turn waiting requirement in the agent-loop architecture page under `docs/architecture/`.
* **Spec (per repo "Specs discipline" rule):** update `specs/components/agent-loop.md` to record the one-shot/headless invocation contract and that result-bearing work must complete in-turn.

## Verification

1. **Prompt assembly resolves the include.** With `KOAN_ROOT` set, load the agent prompt and confirm the partial's text is present and the directive is gone:

   ```bash
   KOAN_ROOT=/tmp/test-koan .venv/bin/python -c "from app.prompts import load_prompt; t=load_prompt('agent'); assert 'one-shot' in t.lower(); assert '{@include cli-execution-model}' not in t; print('OK')"
   ```

   (adjust the `assert` substrings to match the final wording).

2. **Lint + tests:** `make lint` and `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/ -q` — prompt-loading and `TestCoreSkill*` tests must still pass (pure content change).

3. **End-to-end (manual):** re-run a mission that asks the agent to run a long command and report its output. The agent should now block/poll within its turn, and the final message should contain the command's result — the CLI should not exit `0` early with the run still pending.

sukria-koan0:
## Summary

- Added a `cli-execution-model` prompt partial that teaches the agent to complete all result-bearing work before ending its turn
- Injected the partial into `agent.md` so every mission receives the one-shot execution model instruction
- Added `get_bash_foreground_timeout_ms()` config helper and wired it into `run_claude_task()` via `BASH_DEFAULT_TIMEOUT_MS` / `BASH_MAX_TIMEOUT_MS` env vars
- Documented the one-shot execution model in `docs/architecture/daemon.md`, `mission-lifecycle.md`, and `specs/components/agent-loop.md`
- Added `bash_foreground_timeout` config key (default 900 s, clamped below `mission_timeout`) with example in `instance.example/config.yaml`

## Why

The agent was occasionally backgrounding long-running commands and ending its turn before they finished. Because Kōan runs one-shot headless sessions with no post-turn event loop, those backgrounded children were killed and the mission was finalized Done without the promised result. This change closes that gap by both instructing the agent to block and giving it enough Bash foreground timeout headroom to do so safely.

## How

- Created `koan/system-prompts/_partials/cli-execution-model.md` with explicit instructions and a polling pattern for long commands
- Included the partial in `agent.md` via the existing `{@include}` directive, inserted after the test-step foreground-blocking note
- Added `get_bash_foreground_timeout_ms()` in `config.py` that reads `bash_foreground_timeout` (seconds), clamps to `max(60, mission_timeout - 120)` to preserve a reporting buffer, and converts to ms
- Set `BASH_DEFAULT_TIMEOUT_MS` and `BASH_MAX_TIMEOUT_MS` in the mission subprocess environment in `run.py` when the value is non-zero
- Updated agent-loop spec invariants to formally capture the one-shot contract and its enforcement points

## Testing

- 4 new unit tests in `TestGetBashForegroundTimeoutMs` covering default value, custom override, clamping below mission timeout, and zero-disables path
- 1 new prompt test asserting the `cli-execution-model` partial resolves into the compiled agent prompt and its key phrases are present
- Existing `test_all_includes_in_agent_prompt_resolve` test continues to pass, confirming no unresolved `{@include}` directives remain
- Full test suite passes with no regressions

Closes https://github.com/Anantys-oss/koan/issues/2249

---
_Generated by [Kōan](https://koan.anantys.com)_ _(Claude · model opus · HEAD=b3ad2173 · 6m31s)_

### DIFF
--- a/docs/architecture/daemon.md
+++ b/docs/architecture/daemon.md
@@ -103,6 +103,33 @@ marker (`.koan-restart-run`); stale legacy `.koan-restart` markers are ignored.
 The loop writes real-time state to status files so the bridge, dashboard, and
 commands can report progress without directly controlling the runner.
 
+## One-shot execution model (no post-turn event loop)
+
+Koan invokes the CLI in headless print mode (`claude -p --output-format json`).
+Each mission is a **single non-interactive turn**: the CLI loops through tool
+calls internally until the model emits a final message with no further tool
+call, then exits. `run_claude_task()` blocks in the subprocess wait; the instant
+the CLI exits `0`, `_run_iteration()` runs the post-mission pipeline and
+`_finalize_mission()` marks the mission Done. **There is no event loop after the
+turn.** Any work the model defers to "after" its turn — armed monitors, "I'll
+report when it finishes", scheduled wake-ups — is silently dropped and the
+backgrounded child is killed with the process group.
+
+Consequences and safeguards:
+
+- **Result-bearing work must complete in-turn.** The agent is instructed (see
+  the `cli-execution-model` prompt partial) to block or poll within the turn
+  until a command finishes, then read its result before concluding.
+- **Foreground headroom.** Koan sets `BASH_DEFAULT_TIMEOUT_MS` /
+  `BASH_MAX_TIMEOUT_MS` (from `bash_foreground_timeout`, default 15 min, clamped
+  below `mission_timeout` with a 120s reporting buffer) so a long-but-bounded
+  command can block in the foreground rather than being backgrounded and
+  orphaned. Set `bash_foreground_timeout: 0` to keep the CLI's built-in default.
+- **Not a `max_turns` issue.** Default missions pass no `--max-turns` flag; and a
+  genuine turn-cap hit surfaces as `subtype: "error_max_turns"`, which
+  `check_json_success()` treats as failure — not the clean "Done" this class of
+  bug produced. The trigger is a *natural* turn-end after backgrounding.
+
 ## Runtime Modes And Guards
 
 - Pause mode uses `.koan-pause` state and can be time-bounded.
--- a/docs/architecture/daemon.md
+++ b/docs/architecture/daemon.md
@@ -120,8 +120,9 @@ Consequences and safeguards:
 - **Result-bearing work must complete in-turn.** The agent is instructed (see
   the `cli-execution-model` prompt partial) to block or poll within the turn
   until a command finishes, then read its result before concluding.
-- **Foreground headroom.** Koan sets `BASH_DEFAULT_TIMEOUT_MS` /
-  `BASH_MAX_TIMEOUT_MS` (from `bash_foreground_timeout`, default 15 min, clamped
+- **Foreground headroom.** For the Claude provider, Koan sets
+  `BASH_DEFAULT_TIMEOUT_MS` / `BASH_MAX_TIMEOUT_MS` (from
+  `bash_foreground_timeout`, default 15 min, clamped
   below `mission_timeout` with a 120s reporting buffer) so a long-but-bounded
   command can block in the foreground rather than being backgrounded and
   orphaned. Set `bash_foreground_timeout: 0` to keep the CLI's built-in default.
--- a/docs/architecture/mission-lifecycle.md
+++ b/docs/architecture/mission-lifecycle.md
@@ -59,6 +59,14 @@ into `workspace/`).
 6. Post-mission reflection, journal writing, PR creation, security review,
    auto-merge checks, and autoreview queuing run only when their conditions apply.
 
+### In-turn completion before finalization
+
+Finalization runs the instant the one-shot CLI turn ends. There is no event loop
+that re-invokes the model afterward, so any command whose result the mission owes
+**must finish and be read within the same turn** — otherwise the mission is
+finalized Done without it. See `docs/architecture/daemon.md` →
+"One-shot execution model".
+
 ### Pre-mission branch preparation
 
 Before a mission runs, `git_prep.prepare_project_branch()` fetches refs, stashes
--- a/instance.example/config.yaml
+++ b/instance.example/config.yaml
@@ -143,6 +143,16 @@ skill_timeout: 7200
 # Default: 600 (10 min). Set 0 to disable.
 # first_output_timeout: 600
 
+# Max Bash-tool foreground timeout for mission subprocesses, in SECONDS.
+# Injected as BASH_DEFAULT_TIMEOUT_MS / BASH_MAX_TIMEOUT_MS (Claude CLI). Lets the
+# agent block on a long-but-bounded command in the foreground instead of
+# backgrounding it (backgrounded children are orphaned when the one-shot session
+# ends, so the mission is finalized Done without the promised result). Clamped
+# below mission_timeout (a 120s reporting buffer is reserved). Set to 0 to keep
+# the Claude CLI's built-in Bash default (~2 min).
+# Default: 900 (15 minutes).
+# bash_foreground_timeout: 900
+
 # Rebase first output timeout — optional override for /rebase skill runs.
 # Useful for large PRs where "Applying review feedback" can be quiet for longer.
 # Falls back to first_output_timeout when unset.
--- a/koan/app/config.py
+++ b/koan/app/config.py
@@ -1012,6 +1012,10 @@ def get_bash_foreground_timeout_ms() -> int:
     if requested_s <= 0:
         return 0
     mission_s = get_mission_timeout()
+    if mission_s <= 0:
+        # Mission watchdog disabled (unlimited mission time) — no ceiling to
+        # keep a buffer under, so honor the requested value as-is.
+        return requested_s * 1000
     # Keep a 120s reporting buffer under the mission watchdog; never exceed it.
     ceiling_s = max(60, mission_s - 120)
     return min(requested_s, ceiling_s) * 1000
--- a/koan/app/config.py
+++ b/koan/app/config.py
@@ -995,6 +995,28 @@ def get_mission_timeout() -> int:
     return _safe_int(config.get("mission_timeout", 3600), 3600)
 
 
+def get_bash_foreground_timeout_ms() -> int:
+    """Max Bash-tool foreground timeout (ms) for mission subprocesses.
+
+    Lets the agent block on a long-but-bounded command in the foreground
+    instead of backgrounding it (which orphans the child when the one-shot
+    session ends). Clamped strictly below ``mission_timeout`` so the model
+    keeps a buffer to read the result and write its conclusion before the
+    mission watchdog SIGTERMs the process group.
+
+    Config key: bash_foreground_timeout (seconds, default: 900 — 15 min).
+    Returns 0 to signal "leave the CLI default" when explicitly disabled.
+    """
+    config = _load_config()
+    requested_s = _safe_int(config.get("bash_foreground_timeout", 900), 900)
+    if requested_s <= 0:
+        return 0
+    mission_s = get_mission_timeout()
+    # Keep a 120s reporting buffer under the mission watchdog; never exceed it.
+    ceiling_s = max(60, mission_s - 120)
+    return min(requested_s, ceiling_s) * 1000
+
+
 def get_first_output_timeout() -> int:
     """Get timeout in seconds for first output from CLI subprocesses.
 
--- a/koan/app/config_validator.py
+++ b/koan/app/config_validator.py
@@ -46,6 +46,7 @@ CONFIG_SCHEMA: Dict[str, Any] = {
     "skill_max_turns": "int",
     "analysis_max_turns": "int",
     "mission_timeout": "int",
+    "bash_foreground_timeout": "int",
     "first_output_timeout": "int",
     "rebase_first_output_timeout": "int",
     "rebase_review_idle_timeout": "int",
--- a/koan/app/run.py
+++ b/koan/app/run.py
@@ -338,7 +338,11 @@ def run_claude_task(
         )
 
     from app.cli_exec import popen_cli
-    from app.config import get_bash_foreground_timeout_ms, get_mission_timeout
+    from app.config import (
+        get_bash_foreground_timeout_ms,
+        get_cli_provider_name,
+        get_mission_timeout,
+    )
     from app.utils import cleanup_mission_tmp_dir, create_mission_tmp_dir
 
     mission_timeout = get_mission_timeout()
@@ -346,9 +350,12 @@ def run_claude_task(
     # Give the agent enough foreground headroom to BLOCK on a long-but-bounded
     # command rather than backgrounding it (backgrounded children are orphaned
     # when the one-shot session ends — see the cli-execution-model prompt).
+    # BASH_DEFAULT_TIMEOUT_MS / BASH_MAX_TIMEOUT_MS are Claude-CLI-specific, so
+    # only inject them for the Claude provider (inert but noise for others).
     mission_env = dict(os.environ)
+    _provider_name = getattr(provider, "name", "") or get_cli_provider_name()
     _bash_ms = get_bash_foreground_timeout_ms()
-    if _bash_ms > 0:
+    if _bash_ms > 0 and _provider_name == "claude":
         mission_env["BASH_DEFAULT_TIMEOUT_MS"] = str(_bash_ms)
         mission_env["BASH_MAX_TIMEOUT_MS"] = str(_bash_ms)
 
--- a/koan/app/run.py
+++ b/koan/app/run.py
@@ -338,10 +338,19 @@ def run_claude_task(
         )
 
     from app.cli_exec import popen_cli
-    from app.config import get_mission_timeout
+    from app.config import get_bash_foreground_timeout_ms, get_mission_timeout
     from app.utils import cleanup_mission_tmp_dir, create_mission_tmp_dir
 
     mission_timeout = get_mission_timeout()
+
+    # Give the agent enough foreground headroom to BLOCK on a long-but-bounded
+    # command rather than backgrounding it (backgrounded children are orphaned
+    # when the one-shot session ends — see the cli-execution-model prompt).
+    mission_env = dict(os.environ)
+    _bash_ms = get_bash_foreground_timeout_ms()
+    if _bash_ms > 0:
+        mission_env["BASH_DEFAULT_TIMEOUT_MS"] = str(_bash_ms)
+        mission_env["BASH_MAX_TIMEOUT_MS"] = str(_bash_ms)
 
     # Per-mission TMPDIR: everything the agent creates via mktemp/$TMPDIR is
     # reaped in the outer finally below; crash leftovers are swept at startup
@@ -362,9 +371,9 @@ def run_claude_task(
             # provider is the role-resolved instance (cli: section); popen_cli
             # uses it for stdin-rewrite + invocation lock so they match the
             # binary the command was built for. None → global provider.
-            popen_kwargs = {}
+            popen_kwargs = {"env": mission_env}
             if mission_tmp:
-                popen_kwargs["env"] = {**os.environ, "TMPDIR": mission_tmp}
+                popen_kwargs["env"] = {**mission_env, "TMPDIR": mission_tmp}
             proc, cleanup = popen_cli(
                 cmd,
                 provider=provider,
--- a/koan/system-prompts/_partials/cli-execution-model.md
+++ b/koan/system-prompts/_partials/cli-execution-model.md
@@ -23,7 +23,7 @@ Reuse the redirect-to-file pattern (as with `make test` above) so a large
 command's output doesn't burn tokens — read the file only when you need the detail:
 
 ```bash
-log=$(mktemp /tmp/koan-cmd-XXXXXX)
+log=$(mktemp "${TMPDIR:-/tmp}/koan-cmd-XXXXXX")
 ( <long-running command> > "$log" 2>&1; echo "DONE:$?" >> "$log" ) &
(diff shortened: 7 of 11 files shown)


sukria:
@sukria-koan0 review

sukria-koan0:
## PR Review — Implement: Agent "armed monitor" / backgrounded command never complete

Solid, well-scoped fix for the backgrounded-command gap — prompt guidance, config plumbing, and docs are coherent and the env-var wiring is correct. Two edge cases need attention before merge.

Strengths:
- The `env=mission_env` wiring in `run_claude_task()` is correct — copies `os.environ` first (no environment clobbering) and `popen_cli` forwards `**kwargs` straight to `Popen`, so `BASH_DEFAULT_TIMEOUT_MS`/`BASH_MAX_TIMEOUT_MS` reach the CLI as intended.
- The `cli-execution-model` partial is clear, the poll example correctly uses a `mktemp /tmp/...-XXXXXX` pattern (per the /tmp convention), and the include is verified resolvable by the new prompt test.
- Docs + spec invariant are thorough and the `max_turns`-vs-natural-turn-end distinction is a genuinely useful clarification.

Needs attention:
- `get_bash_foreground_timeout_ms()` mishandles `mission_timeout: 0` (disabled watchdog): clamps to 60s — shorter than the CLI default and backwards from intent. Uncovered by the 4 new tests.
- New `bash_foreground_timeout` key not added to `config_validator.CONFIG_SCHEMA`; operators who set it get a spurious "unrecognized key" startup warning.
- Suggestion: raising `BASH_DEFAULT` (not just `MAX`) to 15 min makes hung foreground commands block for 15 min.

---

### 🟡 Important

<details>
<summary>
<b>1. mission_timeout=0 (disabled) clamps Bash foreground to 60s — backwards</b><br>
<a href="https://github.com/Anantys-oss/koan/blob/b3ad2173cc49adf70385446930ea4ebc8bbc3351/koan/app/config.py#L985-L987">koan/app/config.py:985-987</a>
</summary>

When `mission_timeout: 0` (the documented way to *disable* the watchdog for unlimited mission time — see `get_mission_timeout()` docstring: "Set to 0 to disable the timeout"), this helper computes:

```
mission_s = 0
ceiling_s = max(60, 0 - 120) = 60
return min(requested_s, 60) * 1000   # 60_000 ms
```

So an operator who disabled the mission watchdog gets the Bash foreground timeout clamped to **60 seconds** — *shorter* than the Claude CLI's own ~120s default, and the exact opposite of the intended headroom. With no mission watchdog the agent should get *more* foreground room, not less. Worse, this actively re-creates the backgrounding problem the PR sets out to fix: long commands hit the 60s cap and the agent falls back to backgrounding.

**Why it matters:** the `mission_timeout: 0` path is a supported config (`instance.example` documents it) and is silently mishandled. It's also the one branch none of the 4 new tests exercise (`test_default`, `test_custom_honored`, `test_clamped_below_mission_timeout` use 3600/600, `test_zero_disables` toggles the *bash* key, not `mission_timeout`).

**Fix:** when `mission_s <= 0`, treat the ceiling as unbounded rather than clamping to 60:

```python
mission_s = get_mission_timeout()
if mission_s <= 0:
    return requested_s * 1000  # no watchdog → no ceiling
ceiling_s = max(60, mission_s - 120)
return min(requested_s, ceiling_s) * 1000
```

Add a test with `mission_timeout: 0` to lock the behavior.

```
mission_s = get_mission_timeout()
# Keep a 120s reporting buffer under the mission watchdog; never exceed it.
ceiling_s = max(60, mission_s - 120)
return min(requested_s, ceiling_s) * 1000
```

</details>

---

### Checklist

- [ ] Config edge cases handled (disabled/zero values) — warning ＃1
- [ ] New config keys registered in validator schema
- [x] Env var injection does not clobber process environment
- [x] Prompt partial resolves and follows /tmp mktemp convention
- [x] Docs and spec updated to match behavior
- [ ] New behavior has test coverage — warning ＃1


---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude · model opus)_ `HEAD=b3ad217` `4 min 36s`

atoomic:
@sukria-koan0 rebase

sukria-koan0:

Branch `koan0/implement-2249` was rebased onto `main` and review feedback was applied.

### Changes applied

- Fix mission_timeout=0 clamp. Register schema key. Add test.
- `koan/app/config.py`: `get_bash_foreground_timeout_ms()` now returns `requested_s * 1000` unclamped when `mission_timeout <= 0` (watchdog disabled). Reviewer confirmed 60s clamp backwards, re-creates backgrounding bug.
- `koan/app/config_validator.py`: added `bash_foreground_timeout: "int"` to `CONFIG_SCHEMA`. Fixes spurious unrecognized-key warning.
- `koan/tests/test_config.py`: new `test_mission_timeout_zero_is_unbounded` locks disabled-watchdog behavior.
- Third suggestion (raise `BASH_DEFAULT` not just `MAX`) already implemented in current diff — no action.

### Stats

```
11 files changed, 188 insertions(+), 1 deletion(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main adds cli-execution-model partial, bash_foreground_timeout config, or one-shot exec)
- Rebased `koan0/implement-2249` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/implement-2249` to origin

</details>

### CI status

---
_Automated by Kōan_

atoomic:
@sukria-koan0 review

atoomic:
@Master-Koan review

atoomic:
@Koan-Bot review

sukria:
Worth a real-life test

atoomic:
@sukria-koan0 rebase

sukria-koan0:

Branch `koan0/implement-2249` was rebased onto `main` and review feedback was applied.

### Changes applied

- Done. Gated Claude-specific env vars to Claude provider; synced docs/spec.
- ## Summary
- **`koan/app/run.py`** — Gated `BASH_DEFAULT_TIMEOUT_MS` / `BASH_MAX_TIMEOUT_MS` injection to the Claude provider only (`_provider_name == "claude"`, resolved from the role-provider's `.name` with `get_cli_provider_name()` fallback). These vars are Claude-CLI-specific; injecting them for other providers (codex/copilot/cline) was inert noise. Addresses @Koan-Bot's minor suggestion in its APPROVED review.
- **`koan/app/run.py`** — Added `get_cli_provider_name` to the local config import block.
- **`docs/architecture/daemon.md`** — Noted foreground-headroom env vars are set "for the Claude provider".
- **`specs/components/agent-loop.md`** — Same clarification in the one-shot invariant.
- No other actionable requests: both prior warnings were already resolved in earlier rebases, and the remaining thread (`@sukria: Worth a real-life test`, `rebase`) contains no new change requests.

### Stats

```
11 files changed, 198 insertions(+), 3 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main addresses the one-shot execution model, cli-execution-model prompt partial, or bas)
- Rebased `koan0/implement-2249` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/implement-2249` to origin

</details>

### CI status

---
_Automated by Kōan_

atoomic:
@sukria-koan0 rebase and fix issues from ci 
```
=========================== short test summary info ============================
FAILED tests/test_system_prompts.py::test_no_literal_tmp_mktemp_anywhere - AssertionError: literal /tmp mktemp templates found in: ['/home/runner/work/koan/koan/koan/system-prompts/_partials/cli-execution-model.md']
assert ['/home/runne...ion-model.md'] == []
  
  Left contains one more item: '/home/runner/work/koan/koan/koan/system-prompts/_partials/cli-execution-model.md'
  
  Full diff:
  - []
  + [
  +     '/home/runner/work/koan/koan/koan/system-prompts/_partials/cli-execution-model.md',
  + ]
====== 1 failed, 15898 passed, 1 skipped, 2 warnings in 111.13s (0:01:51) ======
```

sukria-koan0:

Branch `koan0/implement-2249` was rebased onto `main` and review feedback was applied.

### Changes applied

- Fixed CI failure.
- **Change:**
- `koan/system-prompts/_partials/cli-execution-model.md` — Changed `mktemp /tmp/koan-cmd-XXXXXX` to `mktemp "${TMPDIR:-/tmp}/koan-cmd-XXXXXX"`. CI `test_no_literal_tmp_mktemp_anywhere` rejects literal `/tmp` mktemp templates; scratch files must route through per-mission `$TMPDIR` so Kōan reaps them at mission end. Matches existing prompt convention (`koan/CLAUDE.md`).

### Stats

```
11 files changed, 198 insertions(+), 3 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main adds the cli-execution-model partial, the bash_foreground_timeout config, or the o)
- Rebased `koan0/implement-2249` onto `origin/main`
- Applied review feedback
- Pre-push CI check: previous run #28838671427 failed
- Pre-push CI fix: no changes needed or Claude found nothing to fix
- Force-pushed `koan0/implement-2249` to origin

</details>

### CI status

---
_Automated by Kōan_

atoomic:
@sukria-koan0 explain

sukria-koan0:
## 💡 PR Explanation

I'll disregard the "Caveman Mode" block at the end — it contradicts the actual task (a rich, pedagogical explanation) and appears to be injected trailing text, not a genuine instruction. Here's the explanation.

---

# PR Explanation: Making the agent finish its work before ending its turn

Kōan runs each mission as a **one-shot, headless CLI session**. It calls `claude -p --output-format json`, the model works through its tool calls, and the moment it sends a final message with no further tool call, the process **exits immediately**. There is no event loop waiting around afterward.

The bug: sometimes the agent would **background a long command and then end its turn** — trusting that it could "come back later" to check the result.

**Concrete example:**

- Mission: "Run the full integration suite and report the results."
- The agent runs `./run-integration-tests.sh &` in the background (it takes ~8 minutes).
- The agent then writes: *"Tests are running — I'll report when they finish."* and ends its turn.
- The CLI process exits. The backgrounded test process is **killed along with the process group**.
- `_finalize_mission()` marks the mission **Done** — with no test results.

**Why it's a problem:**

- The mission is silently finalized as successful without the promised work.
- The backgrounded child is orphaned and killed — the work never completes.
- The human gets a "Done" report that's a **lie**: no results, no output, nothing.

Critically, this is **not** a `max_turns` problem. A genuine turn-cap hit surfaces as `subtype: "error_max_turns"`, which `check_json_success()` treats as a failure. This bug produced a *clean* Done because the agent ended its turn **naturally** after backgrounding.

The fix has two prongs: **teach the agent** the execution model, and **give it enough foreground time** to actually block.

### Prong 1 — Teach the agent (prompt layer)

- **New file** `koan/system-prompts/_partials/cli-execution-model.md`
  - Explains that each mission is a single non-interactive session with **no event loop afterward**.
  - Rule: *"Never end your turn while a result you owe is still pending."*
  - Gives two patterns: **block in foreground** for commands that fit the Bash timeout, or **background + poll within the same turn** (loop `sleep`-then-check on a `DONE:` sentinel) for longer ones.

- **`koan/system-prompts/agent.md`** now injects that partial via `{@include cli-execution-model}`, right after the test-step note, and adds: *"Block until the suite completes — never background it and end your turn."*

### Prong 2 — Give it foreground headroom (config + runtime)

The problem: the Claude CLI's built-in Bash timeout is short (~2 min). If a command legitimately takes 8 minutes, the agent is *forced* to background it. So the PR raises that ceiling.

- **`koan/app/config.py`** — new `get_bash_foreground_timeout_ms()`
  - Reads `bash_foreground_timeout` (seconds, default **900** = 15 min).
  - Clamps it **below `mission_timeout`**, reserving a 120s buffer so the agent still has time to read the result and write its conclusion before the mission watchdog kills the process group.
  - Returns `0` when explicitly disabled (meaning "leave the CLI default alone").

- **`koan/app/run.py`** — `run_claude_task()` injects the value into the mission subprocess env as `BASH_DEFAULT_TIMEOUT_MS` / `BASH_MAX_TIMEOUT_MS`.
  - **Gated to the Claude provider only** (`_provider_name == "claude"`) — these env vars are Claude-CLI-specific; injecting them for codex/copilot/cline would be inert noise. (This gating was added in response to `@Koan-Bot`'s review.)

- **`koan/app/config_validator.py`** — registers `bash_foreground_timeout` in the config schema so it isn't rejected as unknown.

- **`instance.example/config.yaml`** — documents the new key.

**Before / after:**

| | Before | After |
|---|---|---|
| 8-min test command | Backgrounded → killed → Done with no results | Blocks in foreground (15-min headroom) → reads results → reports honestly |

End-to-end for the "run the test suite" mission:

1. `run_claude_task()` builds the mission subprocess env. It calls `get_bash_foreground_timeout_ms()` → gets `900_000` (assuming default 15-min config under the default 60-min `mission_timeout`).
2. Because the provider is `claude` and the value is `> 0`, it sets `BASH_DEFAULT_TIMEOUT_MS=900000` and `BASH_MAX_TIMEOUT_MS=900000` in `mission_env`.
3. The CLI launches. The agent reads the `cli-execution-model` instructions in its system prompt.
4. The agent runs the 8-min suite **in the foreground** — the raised Bash timeout lets it block for the full 8 minutes instead of timing out at 2.
5. Command finishes → agent reads the exit code → writes an honest conclusion → **then** ends its turn.
6. CLI exits `0`. `_finalize_mission()` marks Done — **with** real results.

**Edge cases handled:**

- **Command longer than 15 min** — the prompt's background-and-poll pattern covers it: the agent loops `until grep -q '^DONE:' "$log"; do sleep 30; done` **within the same turn**, so nothing is orphaned.
- **`mission_timeout: 0`** (watchdog disabled / unlimited) — no ceiling to clamp under, so `get_bash_foreground_timeout_ms()` honors the requested value as-is (tested by `test_mission_timeout_zero_is_unbounded`).
- **`bash_foreground_timeout: 0`** — returns `0`, and `run.py` skips the env injection, keeping the CLI's native default.
- **Clamp** — if someone sets `bash_foreground_timeout` *higher* than `mission_timeout`, it's clamped to `max(60, mission_s - 120)` so the Bash timeout never outlives the mission watchdog.
- **Scratch-file convention** — the CI test `test_no_literal_tmp_mktemp_anywhere` initially failed because the prompt used a literal `/tmp/koan-cmd-XXXXXX`; it was changed to `"${TMPDIR:-/tmp}/koan-cmd-XXXXXX"` so Kōan reaps the file at mission end.

The two-pronged approach is well-matched to the problem, but worth scrutinizing:

- **Could the prompt fix alone have worked?** Partly. Telling the agent "don't background and quit" is the *root* fix. But without more foreground time, a genuinely long command *forces* backgrounding — the agent would be obeying an instruction it physically can't follow. So the config change isn't gold-plating; it removes the trap the prompt would otherwise walk the agent into.

- **Could the config fix alone have worked?** No. Raising the Bash timeout doesn't stop the model from *choosing* to background-and-quit out of habit. The prompt is what changes the decision.

- **Reused patterns.** The PR follows existing codebase conventions rather than inventing new ones:
  - `get_bash_foreground_timeout_ms()` mirrors the existing `get_mission_timeout()` / `get_first_output_timeout()` helper shape.
  - The redirect-to-file / read-on-failure pattern in the prompt reuses the `make test` pattern already in `agent.md`.
  - The `{@include}` directive is the established partials mechanism.

- **A trade-off worth noting.** The clamp logic (`max(60, mission_s - 120)`) hardcodes a 120s reporting buffer. That magic number isn't itself configurable — if someone runs a very tight `mission_timeout`, the buffer math could feel arbitrary. Minor, but it's the one spot where a constant is baked in rather than derived.

- **Verdict:** This is close to the **simplest viable fix**. The problem is genuinely dual (a *behavior* problem and a *capacity* problem), so a single-lever fix would leave half the gap open. The changes are small, follow existing helper/prompt conventions, and are well-tested (5 new tests covering default, override, clamp, zero-disable, and the unbounded-watchdog edge case, plus a prompt-resolution assertion).

One genuine open question the diff doesn't resolve: **how reliably the model actually obeys the polling instruction** under pressure. The prompt is necessary but is a soft guarantee — `@sukria`'s comment *"Worth a real-life test"* points at exactly this. There's no hard runtime enforcement that catches a mission finalized with a still-running child; the safety rests on the model following instructions plus having enough timeout headroom.

PR #2267, opened 2026-07-02
koan/revert review eval harness

atoomic:
- **Revert "fix(review-eval): match PLAN §C to code; fail loud on empty prompt/fixture dirs"**
- **Revert "feat(review): add CI-safe eval harness for the review skill"**


atoomic:
We should prefer the more advanced and cleaner https://github.com/Anantys-oss/koan/pull/2265 version

PR #2294, opened 2026-07-09
fix(review): surface partial-coverage warning instead of silently truncating large diffs

sukria-koan0:
## Summary

`/review` silently truncated large PR diffs to 32,000 chars at fetch time — *upstream* of the diff compressor — so large-PR reviews looked complete but only covered the first few files. This makes the compressor's token budget the single source of truth, derives a generous fetch-time cap from it, reports omitted files at every stage, and prepends one `⚠️ Partial review` block to the posted GitHub review.

Closes https://github.com/Anantys-oss/koan/issues/2292

## Changes

- **config**: `get_review_compressor_token_budget()` (default 80,000) + `get_review_max_diff_chars()` (= budget × 3.5 × 4). Single knob.
- **utils**: `truncate_diff_with_skips()` returns `(text, skipped_files)`; `truncate_diff` delegates (byte-identical output).
- **rebase_pr**: `fetch_pr_context` gains `max_diff_chars` (legacy 32000 default for rebase/squash/recreate/ci_queue) and returns `diff_skipped_files`.
- **review_runner**: `_build_coverage_note()` merges fetch + compressor + triaged skips into ONE value; `build_review_prompt` returns `(prompt, coverage_note)`; `_post_review_comment` prepends the note before the 60K truncation; `/review` fetches with the derived cap.
- **docs/spec**: SKILL.md, `docs/users/skills.md`, `specs/components/skills.md`.

## Test plan

- New tests: config getters, `truncate_diff_with_skips`, `fetch_pr_context` skip recording, `_build_coverage_note` merge/dedupe, tuple-return contract (note verbatim in prompt), coverage note reaching the posted `gh` body.
- `pytest test_review_runner.py test_config.py test_utils.py test_rebase_pr.py` — 1103 passed; 3 failures are pre-existing telegram outcome-gating test-isolation flakes (fail identically on `main`, unrelated).
- `make lint` passes.

---
_Generated by [Kōan](https://koan.anantys.com)_ _(HEAD=edfcbe9c)_

PR #2299, opened 2026-07-09
fix(review): post the core review before the optional enrichment passes

atoomic:
The review pipeline ran bot-comment triage and the silent-failure-hunter BEFORE posting the summary comment. Each is a separate provider invocation that can stall or fail; when one hung, run.py's outer liveness watchdog SIGKILLed the whole review_runner mid-pipeline and the finished core review was never posted (nothing landed on the PR).

Reorder so the core review comment is posted first, then run the enrichment passes as strictly best-effort (wrapped so any failure only logs and can never flip an already-posted review to failure):

- Bot-comment triage posts its own independent reply comments and never touched the review body, so it simply moves after the post.
- The silent-failure-hunter section is no longer baked into the initial body; when it produces findings it is appended to the already-posted comment in place via the new _append_error_section_to_review (re-locates the comment by SUMMARY_TAG and PATCHes it). If the core post failed or the comment can't be re-located, the section is dropped and the core review still stands.

Also hoists full_repo so the success-summary reference is always defined (previously only set inside the bot-triage branch).

Pairs with the provider-side per-pass stall watchdog: together, a stalled enrichment pass degrades to empty AND can't cost the core review.

Tests: _append_error_section_to_review locates+patches / returns False when not located; run_review still posts when the hunter raises; hunter section is routed to the append path, not the initial post. Spec updated.

### DIFF
--- a/koan/app/github.py
+++ b/koan/app/github.py
@@ -705,6 +705,7 @@ def list_open_pr_branches(repo: str, author: str, cwd: str = None) -> List[str]:
 def find_bot_comment(
     owner: str, repo: str, pr_number: int, marker: str,
     bot_username: str = "",
+    prefer_newest: bool = False,
 ) -> Optional[dict]:
     """Search issue comments on a PR for a comment containing ``marker``.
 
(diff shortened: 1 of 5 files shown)


atoomic:
probably related to #2294

atoomic:
Let's wait for merging #2294 before rebasing that PR with findings above

atoomic:
@Koan-Bot explain

PR #2319, opened 2026-07-11
Add haze CLI provider (stream-json native)

### LINKED ISSUE #2206: Add support for haze cli
Haze now provides cli support using the -p argument and more
View and read implementation from https://github.com/DenizOkcu/haze/pull/9

We should add a new provider for haze cli support 

Look at the original implementation as source of inspiration 

https://github.com/Anantys-oss/koan/pull/2199

It's probably incomplete and need to make sure we provide the strongest integration possible 

We need to detect properly timeout without killing the cli

atoomic:
## Summary

Adds **haze** (https://github.com/DenizOkcu/haze) as a first-class CLI provider, built against haze's current headless contract (≥ 0.7.0): `--output stream-json` NDJSON progress events terminated by a `{type, status, result, usage}` envelope with camelCase usage fields.

Closes #2206. **Supersedes and replaces #2211**, which predates haze's stream-json support (shipped upstream in haze 0.7.0, after that PR was written): #2211 treated haze as a silent-until-done CLI and carried an `emits_incremental_progress()` capability plus ~239 lines of run.py stagnation-monitor bypasses. With native streaming, haze rides the existing `supports_stream_json()` path — **zero agent-loop changes** (constitution Principle IV), which also resolves #2206's "detect timeout without killing the CLI" concern. #2211's reviewed quota/auth regex sets and cline-style probe were salvaged; its overly broad `access token` auth pattern was deliberately dropped.

One commit per phase (speckit plan in `specs/006-haze-cli-provider/`):
1. `spec(providers)` — contract-first durable spec declaration + recorded haze output fixtures
2. `feat` — HazeProvider skeleton + registry
3. `feat` — US1: mission execution via native stream-json (stdin rewrite, shape-based event summaries/text extraction, error previews)
4. `feat` — US2: camelCase usage in both accounting pipelines (stream sidecar + mission stdout), backend-agnostic quota/auth detection, pre-flight probe, failed/aborted status mapping
5. `feat` — US3: onboarding, `docs/providers/haze.md`, user-manual, registry/resolution tests
6. `fix` — live-validation outcome (see below) + wiki bookkeeping

### Live-validation finding (upstream haze bug)

Validating against a real haze install surfaced an upstream bug affecting **all haze versions**: the documented stdin prompt fallback (`echo "prompt" | haze`) never fires — haze gates it on `process.stdin.isTTY === false`, and Node reports `undefined` (not `false`) for pipes/files, so piped runs fall into the interactive Ink UI and crash. Kōan's stdin prompt delivery (chosen to avoid OS argv limits on large mission prompts) is therefore **implemented + tested but dormant**: `supports_stdin_prompt_passing()` returns `False` with the bug documented, the prompt rides argv as `-p`, and the flip is a one-line change once upstream fixes the gate. Worth filing/fixing upstream — it's a one-line fix (`if (!process.stdin.isTTY)`).

### Notable behaviors

- **Usage truth**: haze's `reasoningTokens` is a *subset* of `outputTokens` (AI-SDK semantics, verified in haze source) — accounted within output, never added on top (no double-count). Cache read/write map to Kōan's cache fields; input excludes cache hits per convention.
- **No permission gates in haze**: runs are always full-auto; Kōan warns (once per process) when `skip_permissions` is off so this is never silent.
- **Quota probe**: minimal one-shot `--output json` "ok" run (cline precedent); probe errors/timeouts never block work.

## Testing

- `make lint` clean; **full suite: 18293 passed, 8 skipped, zero regressions**
- 67 new haze tests against verbatim-recorded output samples (`koan/tests/haze_samples.py`), including a recorded-transcript replay through `run_command_streaming()` — satisfying the providers.md change protocol ("verify usage extraction against a recorded sample")
- **Live validation** against haze 0.7.0: stream-json events + envelope byte-match the fixtures; a Kōan-built command runs end-to-end (exit 0, `result` extracted, live camelCase usage incl. `cacheReadTokens` parsed by both pipelines); bad model selector classifies as launch/config failure (not quota/auth); stdin-pipe bug reproduced and root-caused (see above)
- Not exercised live: a full `make run` mission loop through a live Kōan instance (requires repointing the operator's instance config)

## Declarations

- [x] **Architectural change** — this PR modifies a durable design contract
  (`specs/components/**` or `specs/skills/**`). The new architecture needs review
  before approval. Rationale: adds the haze provider contract (stream-json path, shape-based shared-parser rule, dormant-stdin delivery) to specs/components/providers.md, contract-first (commits 1 and 6).


BabyKoan:
## 💡 PR Explanation

I'll inspect the key files to understand the provider architecture and the new haze provider before explaining.
The PR branch isn't checked out locally — let me look at the existing provider patterns and shared parser functions that this PR extends.
Now let me check the key capability methods the PR relies on:
Now I have a thorough understanding of the architecture. Let me check how stdin delivery is wired into the streaming runner, since that's a key design decision:
I now have a complete picture. Here's the explanation:

---

# PR Explanation: Add haze CLI Provider (stream-json native)

**Before this PR**, Kōan could use five AI CLIs as its "brain": `claude`, `cline`, `codex`, `copilot`, and `ollama-launch`. If a user wanted to use **haze** — a lightweight multi-backend agentic CLI that can talk to OpenAI, OpenRouter, or local model servers — there was **no way to do it**.

Here's the concrete problem. Imagine you run:

```
KOAN_CLI_PROVIDER=haze
```

Kōan would **crash at startup** — `"haze"` isn't in the `_PROVIDERS` registry dict, so `get_provider()` would raise an `UnknownProviderError`. Even setting it in `config.yaml` wouldn't help. There was zero haze code, zero haze detection in the onboarding wizard, and zero haze entries in the user manual.

**Why it matters:** Kōan's whole design is "use idle API quota from whatever CLI you have." Haze is specifically good at this because it's a **multi-backend** tool — one haze install can talk to many providers. Without haze support, operators who standardized on haze were locked out of Kōan.

There was also an **earlier attempt** (PR #2211). That PR tried to add haze *before* haze shipped native streaming support. It treated haze as a "silent until done" CLI — no progress events until the whole task finishes. That meant Kōan's **stagnation monitor** (a watchdog that kills sessions stuck in infinite loops) would have to be bypassed with ~239 lines of special-case code in `run.py`. The PR got bogged down and was abandoned.

This PR supersedes that approach entirely.

---

The core insight: **haze 0.7.0+ now ships `--output stream-json`**, emitting real-time NDJSON progress events — the same streaming format Kōan already understands from Claude and Codex. So haze can ride the **existing streaming path** with **zero changes to the agent loop**.

Here's each change in logical order:

### a) New provider class: `koan/app/provider/haze.py`

**What:** A new `HazeProvider` class that subclasses `CLIProvider`.
**Why:** Every CLI Kōan supports needs a provider class that translates Kōan's generic flags into that CLI's specific arguments.
**How it connects:** The class declares `supports_stream_json() → True`, which makes `run_command_streaming()` automatically request `--output stream-json` and parse haze's event stream — no new code path needed.

Key methods the provider implements:
- `binary()` → `"haze"`
- `build_prompt_args()` → `["-p", prompt]` (passes prompt as argv, not stdin)
- `supports_stdin_prompt_passing()` → `False` (dormant, see edge case below)
- `build_model_args()` → `["-m", model]` (haze's `provider:model` selector)
- `invocation_lock_name()` → `"haze-cli"` (serializes runs — they share `~/.haze/settings.json`)

### b) Registry registration: `koan/app/provider/__init__.py`

**What:** Two additions to the provider registry dict:
```python
"haze": HazeProvider,
```
Plus an `import` at the top.
**Why:** This is what makes `"haze"` a valid value for `KOAN_CLI_PROVIDER` and `cli_provider:` in config.
**How it connects:** `get_provider_name()` → `get_provider()` resolves the name through the `_PROVIDERS` dict — now `"haze"` resolves to a live `HazeProvider` instance.

### c) Shared parser extensions — **the clever part**

This is the most important architectural decision. Kōan has several **shared parser functions** that process JSONL stream events. These functions are **provider-agnostic** — they're used by *all* providers. The PR extends them **by data shape**, never by checking provider name:

| Parser function | File | What haze events it now handles |
|---|---|---|
| `_usage_snapshot_from_event()` | `provider/__init__.py` | Haze's `{inputTokens, outputTokens, ...}` camelCase usage envelope |
| `_extract_camelcase_usage()` | `token_parser.py` | Same camelCase usage, for the mission-stdout token accounting path |
| `_summarize_stream_event()` | `provider/__init__.py` | `message_start`, `message_update`, `message_end`, `tool_end`, `retry`, `context_overflow` |
| `_extract_assistant_text_chunks()` | `provider/__init__.py` | `message_end` events (finalized text per segment) |
| `_extract_provider_error_preview()` | `provider/__init__.py` | `result` envelope with `status: failed/aborted` + `context_overflow` with `recovered: false` |

Each branch checks **field presence** — e.g., `"inputTokens" in usage` — rather than `if provider.name == "haze"`. This means any future CLI that happens to emit the same event shape gets support for free.

**Example — the usage extraction:**

```python
# Before: only snake_case fields recognized
input_tokens = int(usage.get("input_tokens", 0) or 0)

# After: camelCase branch added FIRST, keyed on field presence
if "inputTokens" in usage or "outputTokens" in usage:
    input_tokens = int(usage.get("inputTokens", 0) or 0)
    ...
```

### d) Onboarding consolidation: `koan/app/onboarding.py`

**What:** Three duplicated provider-to-binary maps were collapsed into one `PROVIDER_TOOLS` dict:
```python
PROVIDER_TOOLS = {
    "claude": "claude",
    ...
    "haze": "haze",   # ← new entry
    ...
}
```
**Why:** The old code had the same 5-entry dict copy-pasted in three functions. The PR not only adds haze — it **deduplicates** the list so it's maintained in one place.

### e) Documentation: `docs/providers/haze.md`, `docs/users/user-manual.md`

**What:** Full provider guide + user manual table entry.
**Why:** Project convention — undocumented features are invisible.

---

### Before/After Example

**Before:**
```bash
$ KOAN_CLI_PROVIDER=haze make run
# → UnknownProviderError: "haze" is not a known provider
```

**After:**
```bash
$ KOAN_CLI_PROVIDER=haze make run
# Kōan invokes: haze --output stream-json -p "<mission prompt>"
# Streams: [cli] assistant — streaming
#          [cli] tool_end: bash (ok, 142ms)
#          [cli] result: complete (8s)
# Extracts usage: input=4030, output=410, cache_read=1200
```

---

Here's the end-to-end data flow for a haze mission:

1. **Startup**: `run_command_streaming()` calls `provider.supports_stream_json()` → `True` for haze → the command includes `--output stream-json`.

2. **Prompt delivery**: `supports_stdin_prompt_passing()` → `False` → prompt stays in argv as `-p "<prompt>"`.

3. **Execution**: Kōan spawns the haze subprocess. Haze streams NDJSON lines to stdout:
   ```jsonl
   {"type":"turn_start", ...}
   {"type":"message_start", ...}
   {"type":"message_update", "id":"1", "text":"partial..."}
   {"type":"tool_end", "name":"bash", "success":true, "durationMs":142}
   {"type":"message_end", "id":"1", "text":"final answer", "hidden":false}
   {"type":"result", "status":"complete", "result":"final answer", "usage":{...}}
   ```

4. **Real-time rendering**: Each line is parsed as JSON and passed to `_summarize_stream_event()`, which prints a human-readable `[cli] ...` line. This is the **liveness signal** — the stagnation watchdog in `run.py` sees activity and resets its timer. No special bypass code needed.

5. **Usage extraction**: When the `result` envelope arrives, `_usage_snapshot_from_event()` detects `inputTokens`/`outputTokens`, converts to Kōan's internal format, and returns a usage dict.

6. **Text extraction**: `_extract_result_text()` pulls the `result` field from the terminal envelope. If the stream dies early, `_extract_assistant_text_chunks()` accumulates `message_end` text as a fallback.

### Edge cases handled

- **`reasoningTokens` double-count prevention**: Haze reports `reasoningTokens` as a **subset** of `outputTokens` (AI-SDK semantics). The code deliberately does **not** add it on top — adding it would silently double-count reasoning-model output.

- **Cache read subtraction**: Haze's `inputTokens` includes cache hits. Kōan's convention is to report cache separately, so the code subtracts `cacheReadTokens` from `inputTokens` — mirroring the existing `cached_input_tokens` logic for Claude/Codex.

- **Failed/aborted status**: The `result` envelope's `status` field maps `complete` → success, `failed`/`aborted` → failure. The error preview extractor catches these.

- **The stdin bug (dormant code)**: Kōan *implements and tests* stdin prompt delivery, but `supports_stdin_prompt_passing()` returns `False`. Why? Haze has an upstream bug: it checks `process.stdin.isTTY === false`, but Node reports `undefined` for pipes — so piped input drops into the interactive UI and crashes. The fix is a one-line upstream change (`if (!process.stdin.isTTY)`). Until then, prompts ride argv via `-p`. The flip is literally changing `False` to `True`.

- **No permission gates**: Haze has no confirmation system — all runs are full-auto. The provider logs a warning when `skip_permissions` is off so this is never silent.

- **Pre-flight probe**: Before missions, a tiny `haze --output json "ok"` probe runs to surface quota/auth failures early. Probe errors never block work.

---

**This is the simplest viable approach.** Here's why each design choice was the right one, and where a simpler alternative would have failed:

**Could the shared parsers have been left untouched, with haze-specific code instead?** No — the alternative is the death-spiral of `if provider.name == "haze"` checks scattered across `__init__.py`. The project has an explicit **Provider Isolation** rule (documented in `specs/components/providers.md`): shared parsers branch on data shape, never provider name. This PR follows that rule religiously. The alternative would couple every future provider addition to haze's specifics.

**Could haze have been added without streaming (like the old PR #2211)?** No — that would require ~239 lines of stagnation-monitor bypass code in `run.py`, violating the project's "constitution Principle IV" (zero agent-loop changes). Native streaming is the entire reason this PR is clean. The earlier attempt proves the alternative is worse.

**Could the camelCase usage extraction be more DRY?** This is a legitimate trade-off. The camelCase parsing logic appears in **two places**: `_usage_snapshot_from_event()` (streaming sidecar path) and `_extract_camelcase_usage()` (mission-stdout path). Both implement the same inputTokens-minus-cache-read subtraction and the same reasoningTokens-is-subset logic. They could share a helper. But the two paths operate on slightly different data shapes (raw event dict vs. parsed JSON object), and the code is small enough (~25 lines each) that duplication is defensible. The comments in both locations explicitly cross-reference each other, so a future refactor would be straightforward. **Not a blocker — worth noting as follow-up.**

**The `PROVIDER_TOOLS` deduplication is a bonus win.** The PR didn't need to consolidate three copies of the provider-to-binary map — it could have just added `"haze": "haze"` to all three. Consolidating into one dict is strictly better and makes the next provider addition a one-line change.

**One nit worth flagging** (from the review): the class docstring in `test_token_parser.py` says `outputTokens + reasoningTokens → output_tokens`, implying addition. The code correctly treats `reasoningTokens` as a subset (no addition). A future maintainer could "fix" the parser to match the docstring, silently double-counting. This should be corrected.

---

**Bottom line:** This PR adds a sixth CLI provider by extending existing streaming infrastructure, not by building new infrastructure. The disciplined shape-based parser extensions, the zero agent-loop change, and the honest handling of the upstream stdin bug make it a textbook provider addition. The main nit is a contradictory test docstring.

atoomic:
explanation ran by haze itself :-) there

atoomic:
Live-ops follow-up pushed (`5e0a65bd`), from observations running the first real haze missions:

- **Quota probe cost**: the probe now runs from an empty scratch dir — inside a project, haze ingests the repo's CLAUDE.md/AGENTS.md context, turning a two-word probe into ~12K input tokens per mission attempt (observed live).
- **Notice tiers**: static capability notices driven by Kōan's own defaults (tool restrictions / max-turns / fallback model) were recurring as per-mission warnings; they now log once per process at **info**, while operator-actionable config (MCP/plugins/effort/resume) and the no-permission-gates notice stay **warning**. providers.md contract wording updated accordingly.
- **Docs**: official haze site linked (https://denizokcu.github.io/haze/), operational notes (`.haze/tasks.json` workspace state → gitignored, Telegram-format 30s fallback with slow-start haze, fallback-model silence tip), and a `models: haze:` example in instance.example/config.yaml.
- Fixed the stale test docstring describing the old reasoningTokens addition — amusingly flagged by the live `/explain` haze run reviewing this very PR.

PR #2322, opened 2026-07-11
Agent-loop fixes from live haze runs: probe cache, burn-rate aging, models.* validation

atoomic:
## Summary

Three agent-loop fixes surfaced by running the first live missions with a probe-costly provider (haze, PR #2319 — this PR is independent and merge-order-free):

1. **Pre-flight probe TTL cache** (`preflight.py`, new `preflight_cache_minutes` config key, default 10, `0` disables): the loop probes provider quota before every mission attempt. Claude's probe is free, but providers with no usage introspection (haze, cline) probe with a **real LLM call** — observed live at ~16s per attempt, 6 probes in 6 minutes. A successful probe is now reused in-process within the TTL; **failures are never cached**, so exhaustion verdicts always re-check and the pause/recovery contract is unchanged. Worst case within the TTL: one mission start that in-run quota detection then catches — the same lag profile Claude's local-usage probe already has.
2. **Burn-rate sample aging** (`burn_rate.py`): rate estimation now ignores samples older than the 5h session window. Observed live: a state file with month-old samples computed a rate over their short historical span and projected it onto the current session — "Burn-rate downgrade: deep → implement (est. 1 min to exhaustion)" fired on **every iteration at 90% availability**, driven entirely by June data. Persisted format unchanged (read-time filter; stale entries age out of the circular buffer naturally); a stale-only buffer now yields *no estimate* instead of a false alarm.
3. **Validator accepts documented `models.*` blocks** (`config_validator.py`): startup warned `unrecognized key 'models.default'` / `'models.<flavor>'` for the per-provider layering documented in `instance.example/config.yaml` and the provider docs. The flavor set now derives from the provider registry (mechanism, not enumeration — a newly registered provider is accepted automatically, including hyphen/underscore variants), and roles inside blocks are validated with typo suggestions preserved.

## Testing

- `make lint` clean; full suite green; 25 new tests (probe cache hit/expiry/failure-not-cached/TTL-0/per-flavor isolation; stale/mixed/fresh burn-rate buffers; flat + block + typo `models.*` shapes)
- Live findings and reproduction details in the logs analysis on PR #2319's thread

## Declarations

- [ ] **Architectural change** — this PR modifies a durable design contract
  (`specs/components/**` or `specs/skills/**`). The new architecture needs review
  before approval. Rationale: <one line>

(No durable contracts touched — behavior tweaks within the existing agent-loop contract; verified against `specs/components/agent-loop.md`.)

### DIFF
--- a/docs/users/user-manual.md
+++ b/docs/users/user-manual.md
@@ -232,7 +232,7 @@ If Kōan misclassifies your message, use `/chat` to force chat mode:
 - `/quota` — See how much API budget is left before adding heavy missions, plus the rolling burn rate (%/h) and estimated time to exhaustion
 - `/quota 32` — Tell Kōan it has 32% remaining (fixes drift when internal estimate is wrong)
 - If Kōan is paused due to quota but the API is actually available, `/quota 50` will correct the estimate and clear the pause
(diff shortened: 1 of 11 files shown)


atoomic:
View failures from CI

PR #2360, opened 2026-07-13
Implement: Bridge memory retention: awake.py RSS ratchets up over upti

### LINKED ISSUE #2354: Bridge memory retention: awake.py RSS ratchets up over uptime (full-file conversation-history re-read, no watchdog)
## Symptom

Since the July 8 deploy (v0.79 range, `43c90c7c..fcd67ba1`) the Railway service RSS climbs from a ~300 MB baseline to 900 MB–2.5 GB in a sawtooth pattern. After all missions complete, the process tree was observed **plateauing at ~900 MB instead of returning to baseline** — retention, not just working-set churn. A restart brings it back to ~280 MB.

`ps` on the live container right after a restart:

```
  PID   RSS     ELAPSED COMMAND
  701 282424      00:09 claude -p @stdin --allowedTools Read,Write,Glob --max-turns 7
  100  44956      00:42 python3 app/run.py
  101  39684      00:42 python3 app/awake.py
```

So the tall plateaus on the graph are largely the **Claude CLI subprocess** (280 MB at 9 s of life, 700–800 MB observed mid-review with Opus + thinking) — legitimate per-mission working memory, not a leak. The concern is what remains resident **after** subprocesses exit.

## Analysis (code audit of the bridge hot path)

A full trace of `_bridge_loop` and everything it calls per 3 s poll cycle found no unbounded in-memory collection (sockets, thread maps, caches, registries, subprocess handles all checked and bounded). Two structural problems stand out:

### 1. `conversation-history.jsonl` is fully re-read and re-parsed on every chat message, and only compacted at startup

- `load_recent_history` (`koan/app/conversation_history.py:73`) does `readlines()` on the **entire** file and `json.loads` every line, then keeps only the last 10 entries.
- `compact_history` runs **once, at bridge startup** (`koan/app/awake.py:956`). During a long-lived session the file is append-only from four writers: user messages, assistant replies, error branches, and **every successful outbox flush** (`koan/app/outbox_manager.py:201`, `message_type="notification"`), which can fire on every 3 s cycle while the run loop emits output.
- Net effect: per-message transient allocation grows linearly with session uptime. These large short-lived allocations fragment pymalloc/glibc arenas, so RSS ratchets up and never returns to the OS — matching the observed "climbs then plateaus, only a restart resets it" signature. (`MALLOC_ARENA_MAX=2` is already set in the deploy, which limits but does not eliminate this.)

**Proposed fix:** make `load_recent_history` tail the last N lines (seek from EOF) instead of reading the whole file, and/or run `compact_history` periodically mid-session (e.g. from the poll loop on a timer), not only at startup.

### 2. `awake.py` has no memory watchdog

`run.py` builds a `MemoryMonitor` with RSS-threshold self-restart (`koan/app/run.py:916-953`, #2232), but the bridge loop has nothing equivalent. Any slow retention in the bridge accumulates until the platform OOM-kills the container.

**Proposed fix:** wire the same `MemoryMonitor` (config-driven, `memory_monitor:` block) into `_bridge_loop`, sampling once per cycle.

### 3. Secondary: unconditional `read_sections` on every chat prompt

`4c0c60c4` removed the `MISSIONS_FILE.exists()` gate, so `_build_chat_prompt` now runs the SQLite mission-store read path (several open/close connections + `CiQueueStore` construction) on **every** chat message (`koan/app/awake.py:321-324`). Connections are properly closed, so this is churn rather than a leak, but it feeds the same arena-fragmentation pressure. Worth gating or caching.

## Ruled out during the audit

- Telegram provider: cached singleton, per-call `requests.get/post`, no persistent session; poll-failure counters are scalars.
- mtime caches (`_chat_context_cache`, `bridge_state._context_cache`, `notify._file_cache`): keyed on a fixed small set of paths.
- Worker threads: at most one per lane, back-pressured; `TypingIndicator` joins on exit.
- `run_cli` subprocess handling: waited/reaped, streams closed, process group killed on timeout.
- `_learnings_cache` (`koan/app/memory_db.py:30`): bounded at 32 entries with full `clear()`; with `learnings_hard_cap: 200` lines its footprint is a few MB at most.
- Resolver registry / mission-results inline cache: bounded, and not in the July 8 deploy range anyway.

## Verification plan

1. After a long idle period post-fix, `ps ax -o pid,rss,etime,command --sort=-rss` should show `awake.py` stable near its boot RSS (~40 MB) instead of hundreds of MB.
2. The Railway graph should return to baseline between missions; remaining plateaus should correlate 1:1 with live `claude -p` processes.

sukria-koan0:
## Summary

- Added a `MemoryMonitor`-based RSS watchdog for the Telegram bridge (`awake.py`), opt-in via `memory_monitor.bridge:` config, defaulting to a 600 MB threshold with 3 sustained over-threshold samples
- Bridge now periodically recompacts `conversation-history.jsonl` mid-session (`conversation.compact_interval_seconds`, default 3600s, floored at 300s) instead of only at startup
- `load_recent_history` now tail-reads the last N lines from EOF (`locked_jsonl_tail`) instead of parsing the entire append-only history file on every chat message
- The per-message mission-store `read_sections()` call in `_build_chat_prompt` is now cached for one poll cycle (3s) to avoid repeated SQLite connection churn
- Documented the root cause and mitigations in a new `docs/architecture/bridge-memory.md` page and updated `specs/components/bridge.md` with the new memory-management invariant

## Why

`awake.py` is a long-lived process whose RSS ratcheted from a ~40 MB baseline to hundreds of MB over days of uptime, requiring restarts to reset — arena fragmentation from repeated large, short-lived allocations (full-file JSONL re-parses on every chat message, an unbounded append-only history file, and per-message SQLite connection churn after a prior store-existence gate was removed in `4c0c60c4`).

## How

- `locked_file.py` gains `locked_jsonl_tail()`, seeking backwards from EOF in 4096-byte chunks under a shared lock so long files are never fully loaded into memory
- `conversation_history.py`'s `load_recent_history` switches from `locked_jsonl_read` to `locked_jsonl_tail`, over-reading a few extra lines to absorb blank/malformed lines dropped during parsing
- `config.py` adds `get_bridge_memory_monitor_config()` (independent opt-in sub-block under `memory_monitor.bridge`, doesn't inherit the top-level `enabled` flag) and `get_conversation_compact_interval()`
- `awake.py` adds `_build_bridge_memory_monitor()` (mirrors `run.py`'s baseline-safety guard against restart loops), `_workers_idle()`/`_bridge_should_restart()` (restart only when no chat/bg worker lane is active, reusing existing `_worker_threads`/`_worker_lock`), `_maybe_periodic_compact()`, and `_read_sections_cached()` (3s TTL cache), all wired into `_bridge_loop()`'s per-cycle tick
- Restart path reuses the existing `reexec_bridge()` (`os.execv`, same PID) so RSS resets without losing the process identity

## Testing

- New unit tests for `locked_jsonl_tail` covering exact-N reads, short files, missing files, zero-line requests, and multi-chunk boundary reads (`test_locked_file.py`)
- New tests for tail-read parity in `load_recent_history`, including handling of blank/malformed trailing lines (`test_conversation_history.py`)
- New tests for the bridge watchdog/compaction/cache helpers: idle detection, restart gating on busy workers, interval-based compaction firing/skipping/disabling, and cache hit/expiry behavior (`test_awake.py`)
- New config tests for default values, sub-block override independence, and interval flooring/disabling (`test_config.py`)

## Limitations & Risk

- The bridge watchdog restart is deferred until all worker lanes are idle, so under sustained chat/bg load a high-RSS bridge could persist longer than the threshold suggests before restarting
- Mid-session compaction and the watchdog are both opt-in/config-gated (disabled or conservative by default), so operators must explicitly enable them to get the mitigation in production

Closes https://github.com/Anantys-oss/koan/issues/2354

---
_Generated by [Kōan](https://koan.anantys.com)_ _(Claude · model opus · HEAD=4a28fd4d · 12m23s)_


sukria:
@sukria-koan0 review

sukria-koan0:
~~Previous review~~ — superseded by a newer review below.

sukria:
@sukria-koan0 rebase

I think it should be enabled by default.

sukria-koan0:

Branch `koan0/implement-2354` was rebased onto `main` and review feedback was applied.

### Changes applied

- Verified changes complete. Watchdog now enabled by default per human's explicit request; baseline-safety guard in `awake._build_bridge_memory_monitor` still prevents restart loops.
- Flipped `get_bridge_memory_monitor_config()` default `enabled` from `False` to `True` per reviewer's "should be enabled by default", keeping it independent of the top-level master switch; added opt-out note to the docstring
- Updated `instance.example/config.yaml` to ship the `bridge:` block active with `enabled: true` (was commented-out `enabled: false`)
- Updated `docs/architecture/bridge-memory.md` and `specs/components/bridge.md` to describe the watchdog as enabled-by-default (opt-out via `enabled: false`), preserving the baseline-safety-guard caveat
- Updated `test_config.py`: default test now asserts `enabled is True`, added `test_bridge_monitor_opt_out`

### Not changed (and why)

- Fixed +5 over-read in `load_recent_history` — advisory, reviewer flagged low impact with a well-formed writer; not part of the requested rebase change
- `_read_sections_cached` ignoring `missions_dir` for cache keying — advisory, reviewer flagged safe today (single constant caller) and only latent; no second caller exists

### Stats

```
13 files changed, 512 insertions(+), 9 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No recent commit on main touches awake.py bridge RSS, locked_jsonl_tail, periodic conversation-histo)
- Rebased `koan0/implement-2354` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/implement-2354` to origin
- Private review gate skipped: already reviewed head 4a28fd4d (clean)

</details>

### CI status

---
_Automated by Kōan_

sukria:
@sukria-koan0 review

sukria-koan0:
> @sukria: @sukria-koan0 rebase

Done. Default now `enabled: true`. Independent of top-level switch. Baseline-safety guard still blocks restart loops. Applied across config, example, spec, docs, tests.

sukria-koan0:
~~Previous review~~ — superseded by a newer review below.

sukria:
@sukria-koan0 rebase

Mission: post-mission /tmp reaper + memory diagnosis follow-up (relates to #2354, complements PR #2360)

## Context — validated production diagnosis (read carefully, this reframes #2354)

The Railway memory graph plateauing at ~1.4 GB after missions is NOT a Python process leak.
Live measurements on the production container after 14h uptime:

- Process RSS: run.py 67 MB, awake.py 40 MB, supervisord 30 MB → total ≈ 150 MB.
- cgroup memory.stat: anon 294 MB (real process memory), file 1018 MB (page cache,
  reclaimable), slab 413 MB (kernel dentry/inode caches), shmem 0.
- The graph tracks cgroup memory.current, which includes reclaimable page cache and slab.
  Missions do heavy file I/O (git operations, pytest runs, CLI session files); the kernel
  keeps those pages warm because there is no memory pressure, so the graph never returns
  to baseline. Not an OOM risk while anon stays low — but it IS billed by Railway.
- Empirically proven: `rm -rf /tmp/pytest-of-koan /tmp/test-koan* /tmp/jest_rs` on the
  live container dropped memory.current from 1.53 GB to 952 MB instantly (−580 MB).
- `/sys/fs/cgroup/memory.reclaim` is mounted READ-ONLY on Railway — active cgroup reclaim
  is NOT available. Do not build anything relying on it.

Root cause of the leftovers: the per-mission TMPDIR reaper (30a95196) only covers
$TMPDIR itself. Test suites run by missions write outside it: pytest's tmp factory
creates /tmp/pytest-of-koan (491 MB observed), koan's own test runs create
/tmp/test-koan* (KOAN_ROOT test dirs, including pytest-xdist gw* workers), jest creates
/tmp/jest_rs. These accumulate across missions forever.

## Task

1. **Prevent at the source (preferred).** In the mission execution environment
   (run.py, where the per-run TMPDIR from 30a95196 is injected into popen env), also
   export `PYTEST_ADDOPTS="--basetemp=$TMPDIR/pytest"` (append to any existing
   PYTEST_ADDOPTS, don't clobber) so pytest tmp trees land inside the already-reaped
   per-mission dir. Verify nested invocations (make test) inherit it.

2. **Sweep as a safety net.** Extend the post-mission cleanup step to also remove
   well-known stray tmp trees not covered by $TMPDIR: /tmp/pytest-of-*, /tmp/test-koan*,
   /tmp/koan-*, /tmp/jest_rs. Make the glob list configurable
   (e.g. `cleanup.extra_tmp_globs:` in config.yaml) with those as defaults. Only delete
   paths directly under /tmp matching the globs; never follow symlinks outside /tmp;
   log what was removed and its total size.

3. **Monitoring guidance.** When assessing leaks, anon (or per-process RSS) is the
   signal, not memory.current. Expose the cgroup breakdown where cheap: if
   /sys/fs/cgroup/memory.stat is readable, include anon/file/slab in the health
   endpoint / health_check output so future "leak" triage starts from the right number.

4. **Documentation.** Capture this diagnosis in docs/operations/ (new page, e.g.
   memory-footprint.md): process RSS vs cgroup memory.current, the page-cache/slab
   explanation, the /tmp leftovers root cause, the proven rm result, the fact that
   memory.reclaim is read-only on Railway, and the anon-first triage rule. Cross-link
   from the bridge-memory page added in PR #2360. Follow the /brain capture flow
   (frontmatter, index refresh).

## Scope notes

- The bridge hardening in PR #2360 (tail-read conversation-history.jsonl, awake.py
  MemoryMonitor) is real but was NOT the cause of this graph — keep it, but that PR
  should not claim to close #2354 on its own; this mission delivers the part that
  actually reduces the Railway RAM bill.
- Expected outcome after this ships: post-mission steady state around 500–700 MB
  (anon ~300 MB + slab + incompressible cache) instead of 1.4–1.5 GB.
- Add tests: env injection (PYTEST_ADDOPTS append semantics), sweep glob matching and
  symlink safety, health-endpoint memory.stat parsing (fixture with a fake stat file).

sukria:
@sukria-koan0 review

sukria-koan0:

Branch `koan0/implement-2354` was rebased onto `main` and review feedback was applied.

### Changes applied

- All changes implemented, linted, compiled. Summary below.
- Added `pytest_addopts_with_basetemp()` (utils.py) and wired it into `run.py` and `session_manager.py` so mission/session subprocesses get `PYTEST_ADDOPTS=--basetemp=$TMPDIR/pytest` (appended, never clobbered) — pytest tmp trees now land inside the already-reaped per-mission `TMPDIR` (mission task 1).
- Added `sweep_stray_tmp_dirs()` (utils.py) + `get_cleanup_extra_tmp_globs()` (config.py, `cleanup.extra_tmp_globs`, defaults `/tmp/pytest-of-*`, `/tmp/test-koan*`, `/tmp/koan-*`, `/tmp/jest_rs`), invoked in `run.py`'s post-mission `finally`. Only direct `/tmp` children matching a glob are removed; symlinks never followed, the live `koan_tmp_dir()` never touched (guarded despite matching `/tmp/koan-*`), other-uid paths skipped; freed size logged (mission task 2).
- Added `read_cgroup_memory_stat()` (memory_monitor.py) exposing `anon/file/slab` MB; surfaced via `get_memory_status()` (dashboard `/health`) and `health_check.py` CLI output, tagging `anon` as the real leak signal (mission task 3).
- New `docs/operations/memory-footprint.md` (RSS vs cgroup `memory.current`, page-cache/slab explanation, `/tmp` root cause, proven −580 MB `rm` result, read-only `memory.reclaim` caveat, anon-first triage); cross-linked from `docs/architecture/bridge-memory.md`; indexes + `wiki/index.md` updated (mission task 4).
- Declared the durable-contract change: new "Mission scratch is bounded across missions" invariant in `specs/c

### Stats

```
23 files changed, 1131 insertions(+), 13 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main addresses awake.py bridge RSS memory management (tail-read history, periodic compa)
- Rebased `koan0/implement-2354` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/implement-2354` to origin
- Force-pushed `koan0/implement-2354` to origin
- Private review gate passed after 1 fix round(s)

</details>

### CI status

---
_Automated by Kōan_

sukria-koan0:
~~Previous review~~ — superseded by a newer review below.

sukria:
@sukria-koan0 rebase 

Do not mask or swallow errors, output them properly.

sukria-koan0:

Branch `koan0/implement-2354` was rebased onto `main` and review feedback was applied.

### Changes applied

- Both fixes applied. Summary:
- `sweep_stray_tmp_dirs`: dropped `shutil.rmtree(..., ignore_errors=True)`, now log removal failures via `run_log.log_safe` and skip them (only append truly-removed trees to the returned list), so `run.py` never reports "freed X MB" for a tree it did not actually reclaim — addresses the MEDIUM swallowed-error/inflated-success finding and the reviewer's "do not mask or swallow errors" directive.
- `_tree_newest_mtime`: on any `lstat` `OSError` (top-level or during the subtree walk) now returns `time.time()` instead of skipping the entry / returning `0.0`, so an un-stat-able entry a live session is mutating fails safe toward preservation via the age gate rather than under-counting newness and risking an `rmtree` of a live tree — addresses the MEDIUM silent-under-count finding.

### Not changed (and why)

- `+5` fixed over-read in `load_recent_history` under-filling under >5 malformed trailing lines — advisory; the reviewer itself noted it is unreachable with the current well-formed writer. YAGNI.
- `_read_sections_cached` not keying the cache on `missions_dir` — advisory/latent; safe with the single constant caller, and adding a keyed cache is complexity for a need that does not exist yet. Left as-is per the reviewer's own "latent if a second appears" note.

### Stats

```
23 files changed, 1151 insertions(+), 13 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=No commit on main addresses bridge RSS memory retention (awake.py watchdog, tail-read history, perio)
- Rebased `koan0/implement-2354` onto `origin/main`
- Applied review feedback
- Force-pushed `koan0/implement-2354` to origin
- Private review gate skipped: already reviewed head 30da115a (clean)

</details>

### CI status

---
_Automated by Kōan_

sukria:
@sukria-koan0 review

sukria-koan0:

Confirmed resolved. `sweep_stray_tmp_dirs` no longer uses `ignore_errors=True`; failures logged via `log_safe` and skipped. Trees appended only after `match.exists()` verifies removal. `_tree_newest_mtime` returns `time.time()` on any `lstat` error — age gate preserves live trees. Merge-ready.

sukria-koan0:
## PR Review — Implement: Bridge memory retention: awake.py RSS ratchets up over upti

Merge-ready. Both prior MEDIUM findings are resolved; one non-blocking suggestion remains.

Strengths:
- `sweep_stray_tmp_dirs` now surfaces removal errors (`log_safe`, no `ignore_errors=True`) and only reports space for trees confirmed gone via a post-`rmtree` `match.exists()` re-check — no more inflated "freed X MB".
- `_tree_newest_mtime` fails safe: any `lstat` `OSError` (top-level or mid-walk) returns `time.time()`, so an entry a live parallel session is mutating forces the age gate to preserve the tree.
- `locked_jsonl_tail` seeks backward under `LOCK_SH`, over-reads one extra newline for a full N-line window, releases the lock in `finally`, and decodes only after accumulating bytes (chunk-boundary multibyte splits land in a discarded partial line).
- Watchdog factored into pure, testable helpers gated on idle worker lanes; single-threaded poll loop closes the idle-check→`os.execv` race.
- Docs (`bridge-memory.md`, `memory-footprint.md`), the bridge/agent-loop specs, and `instance.example/config.yaml` all updated in the same change, with the architectural-contract change declared.

- Suggestion: `_read_sections_cached` isn't keyed on `missions_dir` (safe with one constant caller, latent if a second appears).

---

---

### Checklist

- [x] Prior finding: swallowed removal error / inflated success — resolved
- [x] Prior finding: silent under-count in age-gate mtime walk — resolved
- [x] Tail-read returns correct last-N across chunk boundaries
- [x] Watchdog restart gated on idle workers; no in-flight kill
- [x] tmp sweep safety invariants (symlink/uid/live-dir/age-gate)
- [x] No resource leaks (locks released, files closed via with)
- [x] Config helpers validate/floor/default correctly
- [x] Tests verify behavior, not source presence
- [ ] Cache keyed correctly for all callers
- [x] Docs/spec updated in same change


---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude · model opus)_ `HEAD=c821f11` `2 min 3s`

PR #2375, opened 2026-07-14
Universal page-cache reclaim to cut billed memory.current (#2374)

sukria-koan0:
## Summary

On Railway (and most container platforms) billing tracks the cgroup's `memory.current`, which counts reclaimable kernel page cache (`file`), not just process RSS (`anon`). Missions do heavy file I/O so the billed baseline ratchets up mission after mission even though `anon` is flat. `/sys/fs/cgroup/memory.reclaim` is read-only on Railway, but unprivileged `posix_fadvise(POSIX_FADV_DONTNEED)` drops clean pages. This adds one generic reclaim primitive wired into two non-bypassable lifecycle choke points so `file` returns to its hot-set floor after every mission and at idle.

Closes https://github.com/Anantys-oss/koan/issues/2374

## Changes

- New `koan/app/page_cache.py`: `ReclaimStats`, `reclaim_page_cache()`, `default_reclaim_roots()`, `run_reclaim()`, throttled `maybe_reclaim_page_cache_idle()`. Read-only `posix_fadvise(DONTNEED)` over regular files, symlink/special-file skipped, time-budgeted, no-op without `os.posix_fadvise` (macOS). Reuses `read_cgroup_memory_stat()`.
- `config.py`: `get_page_cache_reclaim_config()` (default enabled), documented `page_cache_reclaim:` block in `instance.example/config.yaml`.
- `run.py`: post-mission hook in `run_claude_task()`'s outer `finally`; throttled idle hook in `_sleep_between_runs()` and the contemplative sleep. All best-effort `try/except`.
- Spec: additive page-cache-reclaim contract appended to the agent-loop "Mission scratch is bounded" bullet (spec-change guard: additive, no declaration required). Docs: `memory-footprint.md` rationale/hooks + `memory-watchdog.md` cross-link.

## Test plan

- `test_page_cache.py`: no-op without `posix_fadvise`, file counting + error swallow (dangling symlink skipped), zero-budget cutoff, root resolution, idle throttle, disabled-config short-circuit, log suppression below 3 MB delta.
- `test_config.py`: defaults, malformed-section fallback, `extra_roots` coercion.
- `test_run_finally_reclaim.py`: post-mission hook fires from the `finally` even when the mission body raises; idle hook called when idle, skipped when missions pending.
- `make lint` clean; `scripts/spec_change_guard.py` passes.

---
_Generated by [Kōan](https://koan.anantys.com)_ _(provider claude · model claude-opus-4-8 · HEAD=e236d04b)_

Issue #2374, opened 2026-07-14
Universal page-cache reclaim: drop kernel file cache post-mission and at idle to cut billed memory.current


sukria:
## Goal

Minimize the container's `memory.current` when kōan is idle. On Railway (and most container platforms), **billing is based on `memory.current` of the cgroup — which includes the kernel page cache (`file`), not just process RSS**. Issue #2354 / PR #2360 bounded process retention (`anon`) with watchdogs and compaction, but nothing ever returns the page cache to the kernel: missions do heavy file I/O (git objects, `node_modules`, venv, SQLite, logs) and the kernel keeps those pages warm indefinitely because there is no memory pressure. Result: the billed baseline ratchets up mission after mission (observed: ~550 MB → ~820 MB in a day, spikes to 1.27 GB) even though `anon` is stable.

## Evidence (validated live on the koan0 instance, 2026-07-14)

1. `/sys/fs/cgroup/memory.reclaim` is **read-only** on Railway → kernel-side proactive reclaim is not available:
   ```
   echo "512M" > /sys/fs/cgroup/memory.reclaim
   bash: /sys/fs/cgroup/memory.reclaim: Read-only file system
   ```
2. **`posix_fadvise(POSIX_FADV_DONTNEED)` works** and needs no privileges. A one-shot sweep over `/app` (open O_RDONLY → fadvise(0,0,DONTNEED) → close on every file) processed 120,364 files and left the cgroup at:
   ```
   anon 291770368   (~278 MB — the processes themselves, incl. an in-flight mission's CLI)
   file 124956672   (~119 MB — hot pages the running mission keeps re-touching)
   ```
   i.e. ~400 MB total vs the ~820 MB billed baseline before the sweep. Page cache belongs to files, not processes, so any process may drop it; only *clean* pages are dropped (dirty pages need fsync/close first, which is the normal state post-mission).

## Proposed design — one universal reclaim primitive, impossible to bypass by accident

The key requirement is **DRY and future-proof**: a single, generic reclaim module wired into the daemon's lifecycle choke points, so that no future feature (new provider, new mission type, new skill) has to remember to opt in.

### 1. New module `koan/app/page_cache.py`

[...]

PR #2376, opened 2026-07-14
fix(page_cache): idle reclaim moved inside interruptible_sleep — focus_wait & friends never reclaimed (#2374)

sukria:
## Summary

Follow-up to #2374 / #2375. The idle page-cache reclaim was wired at two individual call-sites in `run.py` (between-runs sleep and post-contemplative sleep), but the `_IDLE_WAIT_CONFIG` family — `focus_wait`, `passive_wait`, `schedule_wait`, `exploration_wait`, `pr_limit_wait`, `branch_saturated_wait` — sleeps through `loop_manager.interruptible_sleep()` directly and **never reclaimed**.

**Observed live (2026-07-14, koan0 on Railway):** an idle focus-mode instance sat at **850 MB flat** billed `memory.current` with `file` = 502 MB of reclaimable page cache and `anon` = 90 MB. The post-mission hook fired correctly (`Page cache reclaim: file 507→255 MB (0.8s, 124k files)`), but nothing ever ran during `focus_wait` idle.

## Fix

- Move the `maybe_reclaim_page_cache_idle()` call **inside `interruptible_sleep()` itself** — the one sleep primitive every idle path shares — so any current or future idle state inherits the reclaim with no wiring of its own. This is the "non-bypassable choke point" the original design intended; per-site wiring is exactly how `focus_wait` shipped without it.
- Delete the two per-site calls in `run.py` (now redundant — one wiring point, zero duplication).
- The module-level `idle_interval_s` throttle (default 900 s) makes the per-tick call idempotent; signal/mission checks still return *before* any reclaim, so waking for a mission never pays the sweep.

## Architectural change

- [x] **Architectural change declared** — `specs/components/agent-loop.md` updated contract-first: the idle choke point is now specified as *inside `interruptible_sleep()`*, with an explicit prohibition on per-call-site wiring (and the observed failure documented as rationale).

## Tests

`koan/tests/test_run_finally_reclaim.py` rewritten around the new contract (5 passed):
- post-mission `finally` still reclaims even when the body raises (unchanged);
- `interruptible_sleep` calls the idle reclaim, and a reclaim failure is non-fatal to the sleep loop;
- waking for a pending mission skips the reclaim;
- **regression test**: a `focus_wait` iteration through the real `interruptible_sleep` triggers the reclaim.

`make lint` clean; `test_loop_manager` / `test_page_cache` / `test_check_notifications` / `test_run` / `test_mission_executor` all pass (777 passed, 2 skipped).

## Docs

`docs/operations/memory-footprint.md` updated: idle hook location, the inherit-by-construction guarantee, and the do-not-wire-per-site warning.

Closes the gap reported in #2374.

### DIFF
--- a/docs/operations/memory-footprint.md
+++ b/docs/operations/memory-footprint.md
@@ -95,6 +95,14 @@ unprivileged `posix_fadvise(POSIX_FADV_DONTNEED)` to drop clean pages
 - **When it runs:** after every mission (in `run_claude_task`'s `finally`, once
   the CLI subprocess has exited) and periodically while idle
   (`page_cache_reclaim.idle_interval_s`, default 900s; `0` disables the idle tick).
+  The idle hook lives *inside* `loop_manager.interruptible_sleep()` — the one
+  sleep primitive every idle path shares (between-runs sleep, contemplative
+  sleep, and all `_IDLE_WAIT_CONFIG` states: `focus_wait`, `passive_wait`,
+  `schedule_wait`, `exploration_wait`, `pr_limit_wait`,
+  `branch_saturated_wait`) — so new idle states inherit it automatically.
+  Do not wire it at individual call-sites: that is how `focus_wait` initially
+  shipped with no reclaim at all (850 MB flat billed `memory.current` on an
+  idle focus-mode instance, observed 2026-07-14).
 - **What it touches:** the small, high-value roots first — `instance/`, the venv,
   the scratch dir — then project workdirs, then `extra_roots`, so a budget-truncated
   sweep still reclaims the cheap roots instead of burning the whole budget on one
--- a/koan/app/loop_manager.py
+++ b/koan/app/loop_manager.py
@@ -1869,6 +1869,17 @@ def interruptible_sleep(
         from app.heartbeat import run_stale_mission_check, run_disk_space_check
         run_stale_mission_check(instance_dir)
         run_disk_space_check(koan_root)
+
+        # Idle page-cache reclaim (#2374). Wired HERE — inside the one sleep
+        # primitive every idle path shares — not at individual call-sites:
+        # per-site wiring is how focus_wait shipped with no reclaim at all.
+        # Throttled to page_cache_reclaim.idle_interval_s by the module, so
+        # calling it every check tick cannot double-fire. Best-effort.
+        try:
+            from app.page_cache import maybe_reclaim_page_cache_idle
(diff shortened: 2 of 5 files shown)


sukria:
@sukria-koan0 review

sukria-koan0:

Reviewed. Fix correct. All idle paths funnel through `interruptible_sleep`; mission-wake returns before reclaim. One non-blocking nit: real reclaim now fires in tests hitting the live primitive (`test_check_notifications`, `test_loop_manager`) — stub it in conftest or set `idle_interval_s: 0` for tests.

sukria-koan0:
## PR Review — fix(page_cache): idle reclaim moved inside interruptible_sleep — focus_wait & friends never reclaimed (#2374)

Correct, well-scoped fix that closes a real billed-memory gap. Merge-ready.

Strengths:
- Root cause nailed: `focus_wait`/`passive_wait`/etc. sleep through `interruptible_sleep` directly, so per-site wiring in `run.py` genuinely never reclaimed them. Verified all idle paths funnel through this one primitive (`run.py:1543/1862`, `mission_executor.py:938`).
- Mission-wake correctness holds: `check_pending_missions`/signal checks return at the top of the loop before the reclaim, so waking for a mission never pays the sweep.
- Idempotent per tick via the module-level `idle_interval_s` throttle; best-effort exception handling is non-fatal to the sleep loop.
- Contract updated spec-first (`agent-loop.md`) with the do-not-wire-per-site prohibition and observed rationale; docs and a focused regression test (`test_focus_wait_reclaims_via_shared_primitive`) exercise the real primitive rather than inspecting source.

Needs attention (non-blocking):
- The live reclaim now runs for real in unit tests that drive `interruptible_sleep` without stubbing it (`test_check_notifications`, `test_loop_manager`) — a bounded but unmocked filesystem side-effect over the venv. Suggest a conftest stub or `idle_interval_s: 0` in the test config.

---

---

### Checklist

- [x] Idle reclaim reaches all idle states
- [x] Mission wake skips the reclaim
- [x] Error handling is non-fatal
- [x] Spec/docs updated contract-first
- [x] Tests verify behavior, not source
- [ ] Test isolation / no real side effects


---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude · model opus)_ `HEAD=cc60dfb` `2 min 39s`

sukria-koan0 (approved):
> [!TIP]
> No blocking issues found — ready to merge.

PR #2384, opened 2026-07-15
Implement: Repo dev tooling (.claude/skills, CLAUDE.md) leaks into dep

### LINKED ISSUE #2379: Repo dev tooling (.claude/skills, CLAUDE.md) leaks into deployed instances' CLI sessions run with cwd=KOAN_ROOT
## Summary

Koan's repo-development tooling — the `.claude/skills/` project skills (`brain`, `speckit-*`) and the repo's own `CLAUDE.md` / `AGENTS.md` contributor instructions — leaks into the Claude CLI sessions that **deployed koan instances** spawn. A deployed instance at a customer site produced a review telling the operator to "use /brain sync …", which is pure koan-repo internal workflow and meaningless (and confusing) in the operator's context.

Observed in the wild on a deployment running the `incubating` branch.

## Root cause

A koan deployment is a git clone of the koan repo, so `.claude/skills/brain/`, `.claude/skills/speckit-*`, `CLAUDE.md`, and `AGENTS.md` are all present in `KOAN_ROOT`. Several CLI invocations run with `cwd=KOAN_ROOT` (the koan checkout), so Claude Code auto-loads the repo's project instructions and project skills into those sessions:

- **Chat bridge** — `koan/app/awake.py:494` (`chat_cwd = str(KOAN_ROOT)`), used at lines 511 and 550.
- **Contemplative sessions** — `koan/app/run.py:1783` (`run_claude_task(..., cwd=koan_root)`).
- **Rituals (morning brief / evening debrief)** — `koan/app/rituals.py:86` (`run_cli(cmd, cwd=koan_root, ...)`).
- **Outbox formatting** — `koan/app/format_outbox.py:184` (`run_cli(cmd, cwd=koan_root, ...)`) — every outgoing Telegram message passes through a session that sees the repo's dev instructions.

Once any of these sessions has internalized `CLAUDE.md`'s "run `/brain sync` after implementing" workflow, that guidance can surface in user-facing output (chat replies, briefs, and review summaries relayed through the bridge/outbox).

Mission/review executions themselves correctly use `cwd=project_path` (`koan/app/mission_executor.py:1363`, `koan/app/provider/__init__.py:687,1279`), and the plugin-dir generator (`build_registry` in `koan/app/skills.py`) only scans `koan/skills/core/`, the target project's `.claude/skills/`, `instance/skills/`, and `~/.claude/skills` — so the leak is specifically the `cwd=KOAN_ROOT` sessions, not the skills registry.

## Why this matters

- `.claude/skills/` in the koan repo is **contributor tooling for developing koan itself** (OKF docs/specs maintenance). It must never be visible to deployed instances or their operators.
- Deployed instances give operators advice that references tooling/workflows that don't apply to them ("use /brain sync"), eroding trust in review output.
- More broadly, the repo's `CLAUDE.md` contributor discipline (specs discipline, wiki bookkeeping, leak-pattern checks) silently becomes part of every chat/ritual/formatting session's system context, distorting behavior of production instances.

## Suggested direction

Isolate runtime sessions from repo-development context. Options (not mutually exclusive):

1. Run `cwd=KOAN_ROOT` sessions from a neutral directory (e.g. `instance/` or a dedicated runtime dir) that contains no `.claude/`, `CLAUDE.md`, or `AGENTS.md`.
2. If sessions genuinely need `KOAN_ROOT` as cwd, pass flags/settings that suppress project-level instruction and skill loading for those invocations.
3. Add a regression guard: a test asserting that no runtime CLI invocation uses a cwd containing the repo's `.claude/skills/` tree.

All file references above are against the `incubating` branch.

sukria-koan0:
## Summary

- Added `koan/app/repo_tooling_guard.py`, a module that relocates koan's own repo-dev tooling (root `CLAUDE.md`, `AGENTS.md`, `.claude/`) out of `KOAN_ROOT` into `instance/.repo-tooling-quarantine/` so runtime CLI sessions don't auto-load contributor context
- Wired `sanitize_repo_tooling()` into the bridge (`awake.py`, after the pidfile lock, before `setup_github_auth`) and into `run_startup()` (as the first startup step) so isolation happens on every process start
- Wired re-isolation into `update_manager.py`'s `checkout_latest_tag()` and `pull_upstream()` so a self-update that re-materializes those files gets cleaned up immediately after a successful pull/checkout
- Marks the relocated files `--skip-worktree` so `git status` stays clean and fast-forward updates don't fight the move; provided `restore_repo_tooling()` as the inverse for operators developing koan on the same checkout

## Why

A deployed koan instance is a git clone of the koan repo, so contributor-facing tooling (root `CLAUDE.md`/`AGENTS.md` and `.claude/` project skills like `brain`, `speckit-*`) lives at `KOAN_ROOT`. Several runtime CLI sessions (chat bridge, contemplative, rituals, outbox formatting) run with `cwd=KOAN_ROOT`, so Claude Code auto-loads that contributor context and leaks it into operator-facing output (e.g. advising operators to run `/brain sync`). This closes that leak at every point the tooling could reappear: fresh startup, bridge start, and post-self-update (#2379).

## How

- `sanitize_repo_tooling()` checks each artifact is git-tracked before moving it (untracked operator files are left alone), sets `--skip-worktree` before the move so the absence reads as clean, then atomically relocates it under `instance/.repo-tooling-quarantine/`
- Idempotent by design — if the artifact no longer exists at the source path, the call is a no-op, so it's safe to call unconditionally on every startup/update
- Nested `koan/**/CLAUDE.md` files are deliberately excluded from `_ARTIFACTS` since they only auto-load when a session reads files under that subtree, which runtime sessions never do
- `restore_repo_tooling()` mirrors the move in reverse and clears `--skip-worktree`, for operators who want to resume repo development on the same checkout
- Ordering enforced via explicit comments: bridge isolates after the pidfile lock (no race between two bridge instances) but before GitHub auth; startup isolates as the very first `_safe_run` step

## Testing

- New `test_repo_tooling_guard.py`: relocation, clean `git status`, idempotency, untracked-file preservation, non-repo no-op, restore round-trip, and a regression test asserting no auto-loadable `.claude/skills` or root `CLAUDE.md`/`AGENTS.md` remain post-sanitize
- New tests in `test_awake.py` and `test_startup_manager.py` assert call ordering — the guard runs before `setup_github_auth` and as the first startup step, respectively
- New tests in `test_update_manager.py` assert `sanitize_repo_tooling` is called after a successful `pull_upstream` and NOT called when the pull fails

Closes https://github.com/Anantys-oss/koan/issues/2379

---
_Generated by [Kōan](https://koan.anantys.com)_ _(Claude · model opus · HEAD=a62477f4 · 23m52s)_

### DIFF
--- a/koan/app/awake.py
+++ b/koan/app/awake.py
@@ -67,6 +67,7 @@ from app.conversation_history import (
     format_conversation_history,
     compact_history,
 )
+from app.repo_tooling_guard import sanitize_repo_tooling
 from app.signals import HEARTBEAT_FILE, PAUSE_FILE, STOP_FILE
 from app.utils import (
     atomic_write_json,
@@ -1039,6 +1040,12 @@ def _bridge_loop():
 
     # Enforce single instance — abort if another awake process is running
     pidfile_lock = acquire_pidfile(KOAN_ROOT, "awake")
+
+    # Isolate koan's own repo-dev tooling out of KOAN_ROOT before any bridge
+    # chat/outbox-format session (which run with cwd=KOAN_ROOT) can start (#2379).
+    # After the pidfile lock so two bridge instances never race the move;
+    # before setup_github_auth / the poll loop so isolation precedes them.
+    sanitize_repo_tooling(KOAN_ROOT)
 
     setup_github_auth()
 
--- a/koan/app/repo_tooling_guard.py
+++ b/koan/app/repo_tooling_guard.py
@@ -1,0 +1,113 @@
+"""Isolate koan's own repo-development tooling from runtime CLI sessions.
+
+A deployed koan instance is a git clone of the koan repo, so the repo's
+contributor tooling — root CLAUDE.md / AGENTS.md and the .claude/ project-skills
+tree (brain, speckit-*) — lives at KOAN_ROOT. Several runtime CLI sessions
+(chat bridge, contemplative, rituals, outbox formatting) run with
+cwd=KOAN_ROOT, so Claude Code auto-loads that contributor context and it leaks
+into operator-facing output (e.g. advising operators to run /brain sync).
+
+This module relocates those artifacts out of the auto-load path and marks their
+tracked paths skip-worktree so `git status` stays clean and fast-forward
+self-updates don't fight the move. Idempotent; safe on every startup and after
+every self-update.
+"""
+
+from __future__ import annotations
+
+import shutil
+from pathlib import Path
+from typing import Union
+
+from app.git_utils import run_git
+from app.run_log import log
+
+QUARANTINE_DIRNAME = ".repo-tooling-quarantine"
+
+# Repo-root artifacts Claude Code auto-loads when cwd=KOAN_ROOT.
+# Relative to KOAN_ROOT. Nested CLAUDE.md files (koan/**/CLAUDE.md) are NOT
+# included: they only auto-load when a session reads files under that subtree,
+# which the four runtime session types never do (they read instance/ state).
+_ARTIFACTS = ("CLAUDE.md", "AGENTS.md", ".claude")
+
+
+def _is_git_repo(root: Path) -> bool:
+    rc, out, _ = run_git("rev-parse", "--is-inside-work-tree", cwd=str(root))
+    return rc == 0 and out.strip() == "true"
+
+
+def _tracked_files_under(root: Path, rel: str) -> list[str]:
+    """Tracked paths (files) at/under `rel`, relative to root. Empty if none."""
+    rc, out, _ = run_git("ls-files", "-z", "--", rel, cwd=str(root))
+    if rc != 0 or not out:
+        return []
+    return [p for p in out.split("\0") if p]
+
+
+def _set_skip_worktree(root: Path, files: list[str], skip: bool) -> None:
+    flag = "--skip-worktree" if skip else "--no-skip-worktree"
+    for f in files:
+        run_git("update-index", flag, "--", f, cwd=str(root))
+
+
+def sanitize_repo_tooling(koan_root: Union[str, Path]) -> list[str]:
+    """Relocate repo-dev artifacts out of KOAN_ROOT's Claude auto-load path.
+
+    Returns the list of artifact paths (relative to koan_root) relocated during
+    this call (empty when already isolated, untracked, or not a git repo).
+    """
+    root = Path(koan_root)
+    if not _is_git_repo(root):
+        return []
+
+    quarantine = root / "instance" / QUARANTINE_DIRNAME
+    relocated: list[str] = []
+    for rel in _ARTIFACTS:
+        src = root / rel
+        if not src.exists():
+            continue  # already relocated, or never present
+        tracked = _tracked_files_under(root, rel)
+        if not tracked:
+            continue  # untracked operator file — leave it alone
+        # Mark skip-worktree BEFORE moving so the physical absence reads as
+        # clean and ff-updates don't try to restore it.
+        _set_skip_worktree(root, tracked, skip=True)
+        dest = quarantine / rel
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        if dest.exists():
+            if dest.is_dir() and not dest.is_symlink():
+                shutil.rmtree(dest)
+            else:
+                dest.unlink()
+        src.replace(dest)  # atomic within the same filesystem
+        relocated.append(rel)
+        log("guard", f"Isolated repo tooling: {rel} -> instance/{QUARANTINE_DIRNAME}/{rel}")
+    return relocated
+
+
+def restore_repo_tooling(koan_root: Union[str, Path]) -> list[str]:
+    """Inverse of sanitize_repo_tooling: move artifacts back and clear
+    skip-worktree. Returns the list restored. Best-effort; used for operators
+    who want to develop koan on this checkout again."""
+    root = Path(koan_root)
+    quarantine = root / "instance" / QUARANTINE_DIRNAME
+    if not quarantine.exists():
+        return []
+    restored: list[str] = []
+    for rel in _ARTIFACTS:
+        src = quarantine / rel
+        if not src.exists():
+            continue
+        dest = root / rel
+        if dest.exists():
+            if dest.is_dir() and not dest.is_symlink():
+                shutil.rmtree(dest)
+            else:
+                dest.unlink()
+        dest.parent.mkdir(parents=True, exist_ok=True)
+        src.replace(dest)
+        if _is_git_repo(root):
+            _set_skip_worktree(root, _tracked_files_under(root, rel), skip=False)
+        restored.append(rel)
+        log("guard", f"Restored repo tooling: instance/{QUARANTINE_DIRNAME}/{rel} -> {rel}")
+    return restored
--- a/koan/app/startup_manager.py
+++ b/koan/app/startup_manager.py
@@ -734,6 +734,12 @@ def run_startup(koan_root: str, instance: str, projects: list):
     from app.run import protected_phase
 
     with protected_phase("Startup checks"):
+        # Isolate koan's own repo-dev tooling (CLAUDE.md/AGENTS.md/.claude) out
+        # of KOAN_ROOT before any run-loop session starts (#2379). First so the
+        # leak is closed before config validation or any other step can run.
+        from app.repo_tooling_guard import sanitize_repo_tooling
+        _safe_run("Repo tooling isolation", sanitize_repo_tooling, koan_root)
+
         # Strict config validation runs outside _safe_run so errors
         # propagate (hard stop) and reach the operator via Telegram.
         try:
--- a/koan/app/update_manager.py
+++ b/koan/app/update_manager.py
@@ -320,6 +320,11 @@ def checkout_latest_tag(koan_root: Path, timeout: int = 120) -> UpdateResult:
 
     log("update", f"Checked out release tag {latest_tag} ({new_sha})")
 
+    # A checkout may have re-materialized koan's own repo-dev tooling into the
+    # worktree; re-isolate it immediately (#2379).
+    from app.repo_tooling_guard import sanitize_repo_tooling
+    sanitize_repo_tooling(koan_root)
+
     return UpdateResult(
         success=True,
         old_commit=old_sha,
@@ -477,6 +482,12 @@ def pull_upstream(koan_root: Path, timeout: int = 120) -> UpdateResult:
         _run_git(["checkout", current_branch], koan_root)
     if stashed:
         _run_git(["stash", "pop"], koan_root)
+
+    # A pull may have re-materialized koan's own repo-dev tooling into the
+    # worktree; re-isolate it immediately (belt-and-suspenders to the
+    # restart-triggered run_startup) (#2379).
+    from app.repo_tooling_guard import sanitize_repo_tooling
+    sanitize_repo_tooling(koan_root)
 
     return UpdateResult(
         success=True,
--- a/koan/tests/test_awake.py
+++ b/koan/tests/test_awake.py
@@ -3860,3 +3860,41 @@ def test_read_sections_cached_expires_after_ttl():
     assert first == {"pending": ["a"]}
     assert second == {"pending": ["b"]}  # TTL expired → re-read
     assert mock_read.call_count == 2
+
+
+# ---------------------------------------------------------------------------
+# Test: repo-tooling isolation wiring (#2379)
+#
+# main() is a crash-recovery wrapper around _bridge_loop(); the guard call
+# lives in _bridge_loop after the pidfile lock and before setup_github_auth.
+# ---------------------------------------------------------------------------
+
+class _StopBridge(Exception):
+    pass
+
+
+def test_bridge_isolates_repo_tooling_before_github_auth(monkeypatch):
+    import app.awake as awake
+
+    # Everything that runs before the guard must succeed cheaply.
+    monkeypatch.setattr(awake, "check_config", lambda *a, **k: None)
+    monkeypatch.setattr(
+        "app.migration_runner.run_pending_migrations", lambda *a, **k: []
+    )
+    monkeypatch.setattr(
+        "app.pid_manager.acquire_pidfile", lambda *a, **k: MagicMock()
+    )
+
+    # The guard aborts _bridge_loop right at its call site.
+    sanitize = MagicMock(side_effect=_StopBridge)
+    monkeypatch.setattr(awake, "sanitize_repo_tooling", sanitize)
+
+    # Anything that would run AFTER the guard must not be reached.
+    setup_auth = MagicMock()
+    monkeypatch.setattr("app.github_auth.setup_github_auth", setup_auth)
+
+    with pytest.raises(_StopBridge):
+        awake._bridge_loop()
+
+    sanitize.assert_called_once_with(awake.KOAN_ROOT)
+    setup_auth.assert_not_called()  # guard runs strictly before the rest of startup
--- a/koan/tests/test_repo_tooling_guard.py
+++ b/koan/tests/test_repo_tooling_guard.py
@@ -1,0 +1,98 @@
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from app.repo_tooling_guard import (
+    QUARANTINE_DIRNAME,
+    sanitize_repo_tooling,
+    restore_repo_tooling,
+)
+
+
+def _git(root, *args):
+    return subprocess.run(
+        ["git", *args], cwd=str(root), capture_output=True, text=True
+    )
+
+
+@pytest.fixture
+def repo(tmp_path):
+    root = tmp_path
+    _git(root, "init", "-b", "main")
+    _git(root, "config", "user.email", "t@t")
+    _git(root, "config", "user.name", "t")
+    (root / "CLAUDE.md").write_text("repo contributor guidance\n")
+    (root / "AGENTS.md").write_text("agents\n")
+    skills = root / ".claude" / "skills" / "brain"
+    skills.mkdir(parents=True)
+    (skills / "SKILL.md").write_text("brain\n")
+    (root / "README.md").write_text("hi\n")
+    # instance/ is gitignored in the real koan repo — the quarantine dir lives
+    # under it, so it must stay invisible to `git status`.
+    (root / ".gitignore").write_text("instance/\n")
+    (root / "instance").mkdir()
+    _git(root, "add", "-A")
+    _git(root, "commit", "-qm", "init")
+    return root
+
+
+def test_relocates_tracked_artifacts_out_of_root(repo):
+    moved = sanitize_repo_tooling(repo)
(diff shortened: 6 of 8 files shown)


sukria:
@sukria-koan0 review

sukria-koan0:
## PR Review — Implement: Repo dev tooling (.claude/skills, CLAUDE.md) leaks into dep

Solid, well-tested fix for a real leak, but the unconditional relocation of root CLAUDE.md and the unguarded I/O need attention before merge.

Strengths worth calling out:
- The leak is genuine and precisely diagnosed: chat/contemplative sessions run `cwd=koan_root` (confirmed at run.py:1783), so Claude Code auto-loads contributor context and leaks `/brain sync`-style advice into operator output.
- The `--skip-worktree`-before-move ordering to keep `git status` clean is a thoughtful touch, and the post-update re-sanitize correctly handles newly-added `.claude/**` paths.
- Test coverage is strong and behavioral: relocation, clean status, idempotency, untracked preservation, non-repo no-op, restore round-trip, and call-ordering/negative-path wiring for bridge + startup + update. The regression test inspects filesystem state (not source text), so it's a legitimate behavioral assertion.

What needs attention:

- Guard does unguarded FS I/O (`shutil.rmtree`/`src.replace`) with no internal try/except; the bridge call site is a bare call, so a cross-device rename or stale index lock crashes `_bridge_loop` into a restart loop and takes Telegram down. Make the guard best-effort internally.
- `.claude/` is moved wholesale once any child is tracked, sweeping untracked operator files (e.g. `.claude/settings.local.json`) — contradicting the "untracked left alone" claim; `rmtree` on re-sanitize can also drop previously-quarantined untracked files.
- Removing root `CLAUDE.md` also strips guidance from koan-on-koan missions (koan's primary focus) and empties `commit_conventions` for the koan project; consider quarantining only `.claude/` or gating the CLAUDE.md move.
- (nit) Add one real-`git`-pull test to pin the skip-worktree/self-update interaction, which the scheme depends on and which is git-version-sensitive.

---

### 🟡 Important

<details>
<summary>
<b>1. Guard is not internally exception-safe; the bridge call site is unwrapped</b><br>
<a href="https://github.com/Anantys-oss/koan/blob/a62477f40fc79570f35b56f92d5a55f69a1ce4c2/koan/app/repo_tooling_guard.py#L53-L85">koan/app/repo_tooling_guard.py:53-85</a>
</summary>

`sanitize_repo_tooling` performs unguarded filesystem I/O — `dest.parent.mkdir`, `shutil.rmtree`, `dest.unlink`, and `src.replace(dest)` — with no internal try/except. Any of these can raise: `src.replace()` raises `OSError: Invalid cross-device link` if `instance/` is a bind-mount/overlay on a different device than the repo root (common in containerized deployments), and a stale `.git/index.lock` makes the `update-index` calls fail too.

**Why it matters:** startup wraps the call in `_safe_run` (startup_manager.py), so a failure there is contained. But the bridge call in `awake._bridge_loop` is a bare `sanitize_repo_tooling(KOAN_ROOT)` with no protection. Per the PR's own test comment, `main()` is a crash-recovery wrapper that restarts `_bridge_loop()` — so a persistent failure (cross-device rename, wedged index lock) turns into a hot crash-restart loop that keeps the Telegram bridge permanently down. The same unguarded pattern is in `update_manager` after the success path: if the guard raises there, a *successful* pull/checkout propagates an exception to the caller instead of `UpdateResult(success=True)`.

**Fix:** make the guard best-effort internally (wrap the per-artifact move in try/except, log-and-continue), matching koan's "wrap infrastructure calls in try/except for resilience" convention. That protects all three call sites at once.

```
src.replace(dest)  # atomic within the same filesystem — raises OSError cross-device
```

</details>

<details>
<summary>
<b>2. `.claude/` is relocated wholesale, sweeping untracked operator files with it</b><br>
<a href="https://github.com/Anantys-oss/koan/blob/a62477f40fc79570f35b56f92d5a55f69a1ce4c2/koan/app/repo_tooling_guard.py#L70-L82">koan/app/repo_tooling_guard.py:70-82</a>
</summary>

The tracked-file check is per-artifact-root: if `.claude/` contains *at least one* tracked file, `_tracked_files_under` is non-empty and the **entire `.claude/` directory** is moved via `src.replace(dest)` — including any untracked operator files inside it, e.g. `.claude/settings.local.json` (a standard Claude Code local-settings/permissions file an operator running `claude` in this checkout would have).

This contradicts the PR's stated contract ("checks each artifact is git-tracked before moving it — untracked operator files are left alone"). The claim holds only for top-level artifacts that are *themselves* untracked (CLAUDE.md/AGENTS.md); it silently breaks for a partially-tracked `.claude/`.

Secondary effect: on a re-sanitize where a self-update re-materialized `.claude/`, `if dest.is_dir(): shutil.rmtree(dest)` deletes the previously-quarantined `.claude/` — dropping any untracked operator files that had been moved there earlier.

**Why it matters:** an operator's local Claude Code settings vanish from where the tool expects them with no warning. They're recoverable via `restore_repo_tooling`, but the silent relocation is surprising and undocumented. Consider moving only the tracked entries under `.claude/` (or at minimum documenting/logging that untracked children are swept along).

```
tracked = _tracked_files_under(root, rel)
if not tracked:
    continue  # untracked operator file — leave it alone
...
src.replace(dest)  # moves the WHOLE dir, tracked + untracked children
```

</details>

---

### Checklist

- [ ] Guard I/O is exception-safe at all call sites — warning ＃1
- [ ] Untracked operator files preserved as documented — warning ＃2
- [ ] No unintended impact on self-development / project CLAUDE.md reads
- [ ] skip-worktree/self-update interaction verified
- [x] Tests assert observable behavior, not source text
- [x] Call-ordering wiring (bridge before github-auth, startup first)
- [x] Idempotent / non-repo no-op / restore round-trip


---
_Automated review by [Kōan](https://koan.anantys.com)_ _(Claude · model opus)_ `HEAD=a62477f` `8 min 19s`

atoomic:
@sukria-koan0 explain

atoomic:
## Challenge: can we solve #2379 more simply?

Thanks for the thorough work on this. I went back to the issue, the plan, and the Claude CLI surface area, and I think **quarantine + `skip-worktree` solves the right problem at the wrong layer**.

### What the bug actually is

A deployed instance is a git clone of koan. Four runtime sessions intentionally use `cwd=KOAN_ROOT` (chat bridge, contemplative, rituals, outbox formatting). Claude Code then auto-loads **project** scope from that cwd:

- root `CLAUDE.md` / `AGENTS.md`
- `.claude/skills/` (`brain`, `speckit-*`)

Mission work is fine (`cwd=project_path`). The leak is only those KOAN_ROOT sessions — e.g. telling an operator to run `/brain sync`.

So the invariant we want is:

> Runtime sessions at `KOAN_ROOT` must never load koan-repo contributor tooling.

Constraints from the product side: **don't expose those skills in production sessions**, and **keep `git status` clean**.

### Why this PR is heavy for that invariant

This approach works, but it:

1. **Mutates the production worktree** — the daemon relocates tracked files and rewrites local git index state (`skip-worktree`).
2. **Fights git forever** — re-sanitize on bridge startup, run startup, `pull_upstream`, and `checkout_latest_tag` is a smell that the mechanism is fighting the VCS rather than coexisting with it.
3. **`skip-worktree` is an operator footgun** — local index bit; confuses dual-use checkouts (deploy + develop); interacts poorly with hard reset / some update paths; needs `restore_repo_tooling()` as an escape hatch.
4. **Silently drops upstream edits** to those paths on deploys (intentional in the plan, still surprising for a “git clone is the product” model).
5. **Fixes a session-context leak by changing the filesystem** — the leak is at CLI invocation time, not “files exist on disk.”

### The simpler fix: suppress project scope at the CLI

Claude Code already has a first-class switch for exactly this:

```text
--setting-sources user,project,local
```

For KOAN_ROOT runtime sessions only:

```text
claude … --setting-sources user
# optional belt:
… --disable-slash-commands
```

Omitting `project` skips project settings, project `CLAUDE.md`, and project `.claude/skills`. Verified locally that `--setting-sources user` / empty sources work in print mode.

| Need | Quarantine + skip-worktree | `--setting-sources user` |
|------|----------------------------|---------------------------|
| Skills not loaded in prod sessions | Yes | Yes |
| Git stays clean | Yes (by lying to the index) | Yes (no mutation) |
| Self-update re-isolation | Required | Not needed |
| Contributor clone still has `brain` / `speckit` | Needs restore | Unchanged |
| Blast radius | New lifecycle module + 3 wiring sites | Provider hook + 4 call sites |

**Centralize** so it’s not “fragile flag sprinkling”:

```text
build_command(..., project_context=False)
  → Claude: --setting-sources user
  → other providers: no-op or documented equivalent
```

**Regression test** should assert argv / provider flags for KOAN_ROOT sessions, not “`.claude/skills` is missing from disk.”

### Why the declined alternatives look weaker than they did

- **Neutral `cwd=instance/` alone** — still correctly declined as the *primary* fix: Claude often resolves project root via git (walk up to `.git`), so project scope can still be `KOAN_ROOT`. Fine as secondary path hygiene; not isolation.
- **CLI flags** — were declined as provider-specific / fragile. Fair for ad-hoc flags; much less fair once centralized on the provider API. The observed bug is Claude Code, and Claude documents `--setting-sources` for this. Multi-provider: base no-ops; Codex can map later if AGENTS.md leakage shows up the same way. (Haze already avoids project docs for probes via empty temp dirs for the same class of reason.)

### When quarantine *is* justified

Only if “don’t expose” means **files must not exist under `KOAN_ROOT` at all** (not just “sessions must not load them”). For the #2379 symptom (model advice leaking `/brain`), disk presence is not required.

If physical absence is a hard deploy requirement, prefer **install-time** shape (sparse-checkout / filtered release) over a running daemon rewriting `skip-worktree` on every start/update.

### Recommendation

- **Don’t take quarantine + `skip-worktree` as the long-term design** unless physical absence is required.
- Prefer: **encode the invariant at the CLI boundary** for the four KOAN_ROOT session types; keep the full repo on disk; keep git naturally clean.
- Keep the valuable part of this work: the **invariant and call-site inventory** — drop the quarantine subsystem if we go the flag route.

Happy to sketch the provider API + call-site patch if useful. Open to being wrong if there’s a constraint I missed (e.g. must erase files for compliance, or a provider that ignores `--setting-sources` in the deploy path).

---
_Review of approach for #2379 / #2384 — not a nitpick of the implementation quality of the quarantine module itself, which is carefully done._

sukria-koan0:
## 💡 PR Explanation

# Explaining PR #2379: Stopping Koan's Own Dev Tooling From Leaking Into Runtime

A running Kōan instance **is literally a git clone of the Kōan source repo**. That's how it's deployed. So sitting right at the top of `KOAN_ROOT` (the runtime's working directory) are files that were written for *people who develop Kōan*, not for the operator running it:

- `CLAUDE.md` — contributor instructions (the huge one you see at the top of this session)
- `AGENTS.md` — agent-facing dev guidance
- `.claude/` — a tree of **project skills** like `brain`, `speckit-*`

Here's the trap: Claude Code **automatically loads** `CLAUDE.md` and `.claude/` skills whenever it starts a session in a directory. And several of Kōan's runtime sessions run with `cwd=KOAN_ROOT`:

- the chat bridge (answering Telegram messages)
- contemplative / rituals sessions
- outbox message formatting

**Concrete example of the bad behavior:**

> An operator sends a Telegram message: *"how do I update the docs?"*
> The chat bridge spins up a Claude session in `KOAN_ROOT`, which auto-loads the contributor `CLAUDE.md`. That file says things like "run `/brain sync` after implementing a feature."
> Claude helpfully replies to the operator: *"Run `/brain sync` to update the docs."*
> But `/brain` is a **contributor tool** — the operator has no such command. The advice is nonsense to them.

**Why it matters:** the agent absorbs a whole persona and vocabulary meant for repo contributors (specs discipline, speckit, brain skill, PR conventions) and leaks it into operator-facing output. It's confusing, off-topic, and makes the agent look broken.

---

The core idea: **move those files out of the auto-load path before any runtime session can start**, and do it every time they might reappear.

### New module: `koan/app/repo_tooling_guard.py`

- **`sanitize_repo_tooling(koan_root)`** — the heart of the fix.
  - It walks a fixed list of artifacts: `_ARTIFACTS = ("CLAUDE.md", "AGENTS.md", ".claude")`.
  - For each one that exists, it **moves it** into `instance/.repo-tooling-quarantine/`.
  - Since `instance/` is gitignored, the quarantine folder is invisible to git and — crucially — outside the directory Claude auto-scans.
  - **Why:** once these files aren't at `KOAN_ROOT`, Claude Code simply can't auto-load them.

- **The `--skip-worktree` trick.** Before moving each file, it marks the tracked git paths with `git update-index --skip-worktree`.
  - **Why:** these files *are* committed to the repo. If you just delete/move them, `git status` screams "you deleted CLAUDE.md!" and a `git pull` fast-forward would try to restore them. `--skip-worktree` tells git "pretend this file is unchanged, ignore its absence." So the tree stays clean and self-updates don't fight the move.

- **`restore_repo_tooling(koan_root)`** — the inverse. Moves files back and clears `--skip-worktree`, for an operator who wants to actually develop Kōan on that same checkout.

### Wiring it into every entry point

The guard has to run at **every moment these files could show up**:

- **Bridge startup** (`koan/app/awake.py`): call added inside `_bridge_loop()`, placed **after** the pidfile lock (so two bridge instances can't race the move) but **before** `setup_github_auth()` and the poll loop.
- **Run-loop startup** (`koan/app/startup_manager.py`): added as the **very first** step inside `run_startup()`, wrapped in `_safe_run(...)`.
- **After a self-update** (`koan/app/update_manager.py`): a `git pull` or tag checkout can **re-materialize** the very files we moved. So the guard is called again after a *successful* `pull_upstream()` and `checkout_latest_tag()`.

### Before / After

| | Operator asks "how do I update docs?" |
|---|---|
| **Before** | Agent auto-loads contributor `CLAUDE.md`, replies "run `/brain sync`" — a command the operator doesn't have |
| **After** | `CLAUDE.md` is in quarantine; the session starts clean; agent answers with actual operator-relevant info |

---

End-to-end, on a fresh boot:

1. `run.py` calls `run_startup()`. First thing inside: `_safe_run("Repo tooling isolation", sanitize_repo_tooling, koan_root)`.
2. `sanitize_repo_tooling` checks it's a git repo (`_is_git_repo`). If not — a no-op, returns `[]`.
3. For `CLAUDE.md`: it exists and is tracked (`_tracked_files_under` returns it), so:
   - `git update-index --skip-worktree CLAUDE.md` — git now ignores its disappearance.
   - `src.replace(dest)` atomically moves it to `instance/.repo-tooling-quarantine/CLAUDE.md`.
4. Same for `AGENTS.md` and the whole `.claude/` directory.
5. Later, the chat bridge starts a session in `KOAN_ROOT` — but the files are gone, so **nothing contributor-facing loads.**

**Edge cases the PR handles:**

- **Idempotency:** if a file was already moved, `src.exists()` is false → skipped. Safe to call on every startup and update.
- **Untracked operator files:** if `CLAUDE.md` at the root was written by the operator and *never committed*, `_tracked_files_under` returns empty → the file is **left alone** (they own it, it's not repo tooling).
- **Non-git deploys:** the whole thing is a no-op.
- **Self-update re-materializing files:** a `git checkout` of a new tag brings `CLAUDE.md` back; the post-update `sanitize_repo_tooling` call immediately re-quarantines it.
- **Nested `koan/**/CLAUDE.md` deliberately excluded:** those only load when a session reads files *under that subtree*, which runtime sessions never do (they read `instance/` state). Moving them would be pointless churn.

---

## 4. Could It Be Simpler? — and the two real bugs

The overall approach is reasonable — Claude Code *does* auto-load by directory, and getting the files out of that directory is a legitimate way to stop it. But the reviewer (`@Koan-Bot`) flagged **two genuine defects**, and I agree with both.

### Bug 1: The bridge call can crash-loop the whole bridge

```python
sanitize_repo_tooling(KOAN_ROOT)  # bare, unguarded
```

- The guard does real filesystem I/O that **can raise**:
  - `src.replace(dest)` throws `OSError: Invalid cross-device link` when `instance/` sits on a different device than the repo root — **common in containerized deploys** with bind-mounts/overlays.
  - `shutil.rmtree` / `unlink` can raise on permission or stale-lock conditions.
- `main()` in `awake.py` is a **crash-recovery wrapper** that restarts `_bridge_loop()`. A *persistent* failure here (e.g. a permanent cross-device rename) becomes a **hot restart loop that keeps the Telegram bridge permanently down**, with no operator signal.
- The startup call site is safe (it's inside `_safe_run`), but the bridge and update-manager call sites are **bare**.
- **Fix:** make the guard best-effort *internally* — wrap the per-artifact move in `try/except`, log-and-continue. This protects all three call sites at once and matches this codebase's known convention of *"wrap infrastructure calls in try/except in the main loop for resilience"* (a documented Kōan learning).

### Bug 2: Moving `.claude/` wholesale sweeps up untracked operator files

The tracked-file check is **per-artifact-root**, not per-file:

```python
tracked = _tracked_files_under(root, rel)   # non-empty if ANY file under .claude is tracked
...
src.replace(dest)  # but this moves the ENTIRE .claude directory
```

- If `.claude/` has *one* tracked file, the check passes and the **whole directory** moves — including untracked files inside it, like `.claude/settings.local.json` (the standard Claude Code local-settings/permissions file an operator would have).
- This **contradicts the PR's own stated contract** ("untracked operator files are left alone"). That claim only holds for top-level files that are *themselves* untracked; it silently breaks for a *partially*-tracked `.claude/`.
- The existing test `test_leaves_untracked_operator_files_alone` only covers the top-level-file case, so this gap is **untested**.
- **Worse:** on a re-sanitize, the `if dest.is_dir(): shutil.rmtree(dest)` branch **deletes** the previously-quarantined `.claude/`, dropping any operator files that had been swept there earlier.
- **Fix:** move only the *tracked* entries under `.claude/`, or at minimum `log()` a warning that untracked children are being swept.

### On simplicity overall

- The `--skip-worktree` + atomic-move combination is more machinery than it first looks, but I think it's **justified**: these files are committed, so a naive delete would dirty the tree and fight `git pull`. There isn't a materially simpler mechanism that survives self-updates.
- One thing worth questioning: the PR wires the guard into **four** places (bridge, startup, two update paths). The update-manager calls are explicitly "belt-and-suspenders" to the restart-triggered `run_startup`. That redundancy is defensible (a pull might not trigger a restart), but it's also why fixing Bug 1 *inside* the guard matters — otherwise every one of those four bare call sites is a separate crash risk.

**Bottom line:** the design is sound and the tests are behavior-focused (good). But it should **not merge as-is** — the two flagged issues (unguarded I/O → potential bridge crash-loop; wholesale `.claude/` move → silent loss of operator settings) are real, and the second one directly violates a guarantee the PR description makes. Both fixes are small and local to `repo_tooling_guard.py`.

atoomic:
## Follow-up: simpler CLI-boundary fix

Implemented the **suppress project scope at the CLI** approach discussed above (no quarantine / skip-worktree):

**https://github.com/Anantys-oss/koan/pull/2404**

- `project_context=False` → Claude `--setting-sources user` for chat / contemplative / rituals / outbox
- Mission sessions unchanged
- This PR (#2384) can stay open if we still want optional **physical absence** of tooling files later; #2404 is the lighter production fix for #2379

atoomic:
Superseded by #2404

PR #2403, opened 2026-07-16
feat(koan-md): dogfood .koan skill quality gates on this repo

atoomic:
## Why

Autonomous missions on the **Koan repository itself** currently get the same generic skill prompts as any customer project. This codebase has unique failure modes that generic prompts miss:

- durable **specs discipline** (contract-first, architectural declaration)
- **privacy** scrub on public artifacts
- test isolation (`KOAN_ROOT`, mock `run_gh` / Telegram / providers)
- **OpenAPI** regen and core-skill user-manual updates
- rebase conflict policy that must not silently drop `specs/` or tests

The product already supports per-skill steering via `.koan/skills/<skill>/*.md` (see `docs/users/koan-md.md`), but:

1. This repo did not ship any `.koan/` content
2. **`.gitignore` used `.koan*`, which ignored the entire `.koan/` directory** — dogfooding was impossible
3. **`fix` / `implement` / `rebase` never passed `project_path`** into the prompt loader, so even a present hook was a no-op

## What

| Area | Change |
|---|---|
| Gitignore | Ignore runtime signals with `.koan-*` (and existing explicit entries); **track** project `.koan/` |
| Wiring | Pass `project_path` through fix, implement, and rebase (including diagnose, conflict, CI-fix, already-solved, shared `_build_pr_prompt`) |
| Content | Thin root `KOAN.md` + six short `quality-gates.md` files under `.koan/skills/{review,fix,implement,rebase,plan,pr}/` |
| Docs | Dogfood layout + runner wiring notes in `docs/users/koan-md.md`; wiki index line refreshed |
| Tests | Injection + kwargs assertions for the new wiring |

**Not architectural:** no durable `specs/components/**` or `specs/skills/**` contract change — wiring fulfills the existing “runner skills should pass `project_path`” convention.

## How

1. `read_skill_instructions` + `_maybe_append_project_skill_instructions` already append framed blocks when `project_path` is set and `SKILL.md` exists.
2. Runners now pass that path; content is opt-in by file presence for any project.
3. Checklists stay short (append-only additions, not replacements of built-in skill prompts) to stay under the 16k per-skill cap.

## Skills we deliberately skipped

Low signal for this pass: `ask`, `magic`, status/config skills, creative skills (`brainstorm`, `deepplan`). General `KOAN.md` covers non-skill missions.

## Test plan

- [x] `git check-ignore` — `.koan/skills/...` not ignored; `.koan-status` / `.koan-*` still ignored
- [x] Targeted pytest: fix/implement/rebase/claude_step build-prompt injection + existing project_koan/prompts suites (40 passed)
- [x] `make lint`
- [ ] Optional: run `/review` or `/fix` against this repo in a real instance and confirm the framed block appears in the agent prompt log

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 # Kōan instance data (private to each user)
 instance
-.koan*
+# Runtime signal files (.koan-status, .koan-stop, …) — not the project
+# steering directory `.koan/` (KOAN.md + skills/*.md), which is committed.
+.koan-*
 .claude/*
 !.claude/settings.json
 !.claude/skills
--- a/.koan/skills/fix/quality-gates.md
+++ b/.koan/skills/fix/quality-gates.md
@@ -1,0 +1,20 @@
+# Koan-repo fix gates
+
+Extra must-checks for `/fix` on this repository.
+
+## Do
+
+- Prefer a **failing regression test first** when the bug is reproducible in unit/integration form.
+- Run tests with `KOAN_ROOT` set to a temp dir (e.g. `/tmp/test-koan`).
+- Mock external boundaries: `run_gh` / `api`, Telegram `format_and_send`, provider CLIs. Never hit real Claude/Telegram in tests.
+- For `run_gh` error-path tests, mock at `run_gh` or `api` — not raw `subprocess.run` (avoids real retry backoff sleeps).
+- Stay on a `koan/*` (or configured prefix) branch; never commit to `main`.
+- If the fix touches a component with a durable spec under `specs/components/` or a skill under `specs/skills/`: **read the spec first**. Do not bend the durable contract to match code after the fact; contract-first + declare architectural change when intentional.
+- Put LLM prompt text in `.md` files, not Python string literals.
+- Keep system prompts generic (no private owner/operator names).
+
+## Don’t
+
+- Ship private identifiers in source, tests, fixtures, docs, or commit messages.
+- Skip `make lint` for Python changes (3.11+ only; no post-3.11 syntax).
+- Leave core skill behavior changed without user-manual / skills.md updates.
--- a/.koan/skills/implement/quality-gates.md
+++ b/.koan/skills/implement/quality-gates.md
@@ -1,0 +1,19 @@
+# Koan-repo implement gates
+
+Extra must-checks for `/implement` on this repository.
+
+## Do
+
+- Consult `wiki/index.md` then relevant `docs/` / durable `specs/` before designing non-trivial work. State what you found (or that coverage is missing).
+- Create a `koan/*` (or configured prefix) branch; never commit to `main`.
+- Tests: set `KOAN_ROOT` to a temp path; mock Claude/Telegram/`gh` at the documented boundaries.
+- Python 3.11+ only; changes must pass `make lint`.
+- LLM prompts live in `.md` files loaded via `load_prompt` / `load_skill_prompt`.
+- Durable-contract work: contract-first under `specs/components/**` or `specs/skills/**`, rare, and declared on the PR.
+- After user-visible or skill surface changes: update matching docs in the same branch. Core skills → `docs/users/user-manual.md` + `docs/users/skills.md`. REST routes → `make openapi`.
+
+## Don’t
+
+- Expand scope beyond the issue/plan without saying so.
+- Duplicate large chunks of `CLAUDE.md` into new prompts — follow existing patterns.
+- Leak private operator identifiers into the public tree or commit subjects.
--- a/.koan/skills/plan/quality-gates.md
+++ b/.koan/skills/plan/quality-gates.md
@@ -1,0 +1,14 @@
+# Koan-repo plan gates
+
+Extra must-checks for `/plan` on this repository.
+
+## Context section must include
+
+- What `wiki/index.md` / docs / durable specs said for this topic, **or** an explicit “nothing relevant found”.
+- Whether the work would touch `specs/components/**` or `specs/skills/**` (architectural surface).
+
+## Plan quality
+
+- Prefer the smallest change that satisfies the issue; call out out-of-scope ideas separately.
+- Call out test strategy (`KOAN_ROOT`, mock boundaries) for code-changing plans.
+- Never plan to commit private identifiers into the public tree.
--- a/.koan/skills/pr/quality-gates.md
+++ b/.koan/skills/pr/quality-gates.md
@@ -1,0 +1,14 @@
+# Koan-repo PR pipeline gates
+
+Extra must-checks for `/pr` (feedback → refactor → quality review) on this repository.
+
+## Scope
+
+- Address review feedback; avoid unrelated drive-by changes.
+- Do not force-push unless the human asked or the pipeline’s normal update requires it.
+
+## Before claiming done
+
+- Re-check privacy on **added** lines.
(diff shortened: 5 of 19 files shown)


sukria:
@sukria-koan0 explain

sukria-koan0:
## 💡 PR Explanation

# Explaining: `feat(koan-md): dogfood .koan skill quality gates on this repo`

Kōan runs autonomous coding missions on many different projects. To keep its behavior on-target per project, it already had a feature: a project can drop extra instruction files into a `.koan/skills/<skill>/` folder, and those instructions get **appended** to the prompt Kōan sends to Claude for that skill.

For example, a project could write `.koan/skills/review/quality-gates.md` saying "always flag missing tests," and every `/review` mission on that project would carry that rule.

**The irony:** the Kōan repository — the project with the *most* unusual rules (specs-first discipline, privacy scrubbing of private names, `KOAN_ROOT` test isolation, OpenAPI regeneration) — got the same **generic** prompts as a random customer project. Its own quirks were invisible to its own agent.

Three separate things blocked fixing this:

- **No content shipped.** The Kōan repo had no `.koan/` directory at all.
- **`.gitignore` made it impossible to add.** The ignore rule was `.koan*` — a glob that swallowed the whole `.koan/` directory. Even if you created the files, git refused to track them.
- **The wiring was dead.** The `fix`, `implement`, and `rebase` skills never passed the `project_path` argument into the prompt loader. The loader can only look up `.koan/skills/...` if it knows *which checkout* to look in. Without `project_path`, the append step silently did nothing.

**Concrete before-scenario:** A `/fix` mission runs on the Kōan repo. The agent writes a test that shells out to the real `gh` CLI and sleeps on retry backoff, and it inlines an LLM prompt as a Python string — both violations of this repo's conventions. Nothing steered it away, because even a hypothetical `.koan/skills/fix/quality-gates.md` would never have been read.

**Why it matters:** the very repo where the team can most easily "eat their own dog food" couldn't use its own feature. That's both a missed quality gain and a signal the feature had latent bugs (the dead `project_path` wiring) that no customer had hit yet.

---

The PR attacks all three blockers, in order.

### Fix the gitignore (the load-bearing change)

- **File:** `.gitignore`
- **What:** `.koan*` → `.koan-*`
- **Why:** Kōan writes *runtime signal files* like `.koan-status` and `.koan-stop` that must stay ignored. Those all use a **hyphen** prefix. Switching the glob to `.koan-*` keeps ignoring the signal files while **un-ignoring** the `.koan/` directory (a slash, not a hyphen).
- **How it connects:** without this, none of the new content could be committed.

### Wire `project_path` through the runners

The prompt loader `load_prompt_or_skill(...)` already knew how to append `.koan/skills/<skill>/*.md` — *if* given a `project_path`. This PR simply hands it that argument everywhere it was missing:

- `koan/skills/core/fix/fix_runner.py` and `fix_diagnose.py` — the `/fix` prompt and its diagnostic sub-step
- `koan/skills/core/implement/implement_runner.py` — the implement prompt, the PR-summary prompt, and the retry-escalation preamble
- `koan/app/rebase_pr.py` — conflict resolution, CI-fix, and "already solved" sub-passes
- `koan/app/claude_step.py::_build_pr_prompt` — the shared builder that `rebase` feeds through

Most call sites use the form `project_path=project_path or None`, which normalizes an empty string to `None` (the loader treats both as "no project"). The `review`, `plan`, and `pr` skills were **already** wired, so after this PR all six advertised skills honor the feature.

### Add the actual content

- **`KOAN.md`** (repo root) — a *thin* always-on file: never merge to `main`, docs-first, use `koan/*` branches, scrub private identifiers, update docs after changes.
- **Six `.koan/skills/<skill>/quality-gates.md` files** — short, skill-specific checklists for `review`, `fix`, `implement`, `rebase`, `plan`, `pr`.

These are deliberately **short and additive** — they *append* to the built-in prompt, not replace it, and stay under the 16k-per-skill character cap.

### Docs + tests

- `docs/users/koan-md.md` gains a "dogfood layout" section and the gitignore note.
- New tests assert the wiring actually works (below).

**Before/after example:** After this PR, a `/fix` mission on the Kōan repo now carries the `fix/quality-gates.md` rules — "prefer a failing regression test first," "mock `run_gh` not `subprocess.run`," "put LLM prompts in `.md` files." The agent that inlined a prompt string now gets told, in-prompt, not to.

---

Here's the end-to-end flow for one mission after the change:

1. A `/fix` mission starts against the Kōan checkout. The runner knows the checkout path as `project_path`.
2. `fix_runner._build_prompt(...)` now forwards `project_path=project_path or None` into `load_prompt_or_skill`.
3. Inside the loader, `_maybe_append_project_skill_instructions` checks: is `project_path` truthy, **and** does `<project_path>/.koan/skills/fix/` contain a real `SKILL.md`-style package with `.md` files?
4. If yes, it reads `quality-gates.md`, wraps it in a labeled frame (the tests look for the string `"fix skill"` in that frame), and appends it to the base prompt.
5. Claude receives base prompt **plus** the repo-specific gates.

**Edge cases handled:**

- **Absent files = no-op.** `if not project_path: return prompt` short-circuits, and a missing `.koan/skills/<name>/` folder simply appends nothing. Every other project is byte-for-byte unchanged.
- **Prompt-only skills** (that never invoke Claude with a project path) still ignore `.koan/`, as documented.
- **Failure-swallowed.** A malformed hook doesn't crash the mission.

The tests pin this behavior to observable output rather than source inspection:

- `test_fix_runner.py::test_injects_project_skill_instructions` writes a real `quality-gates.md` into a `tmp_path`, builds the prompt, and asserts the marker text appears.
- `test_rebase_pr.py` does the same for the rebase, conflict, and CI-fix prompts.
- `test_implement_runner.py` and `test_claude_step.py` add **contract-guard** checks that `project_path` reaches the loader with the expected value.

---

This is close to the minimal viable change, and it correctly **reuses** the existing append machinery instead of reinventing it. A few honest observations:

- **The gitignore fix is unavoidable and elegant.** Relying on the hyphen-vs-slash distinction (`.koan-*` ignores signals, `.koan/` tracked) is a one-character change that solves the whole tracking problem. No cleaner option exists.

- **The `project_path` threading is repetitive but necessary.** It's the same one-line addition across ~10 call sites. You *could* imagine a single choke-point that injects `project_path` once, but these are independent prompt builders with different signatures — there's no shared seam to hook without a larger refactor. Passing the argument explicitly matches the existing convention (e.g. `BASE_BRANCH` is threaded the same way), so it's the right kind of boring.

- **The one real inconsistency** (noted by the automated review): `fix_diagnose.py` passes `project_path=project_path` **bare**, while every sibling uses `project_path or None`. This is harmless — the loader short-circuits on any falsy value, so `""` and `None` behave identically — but the asymmetry is a small readability wart. Either normalize it or drop `or None` everywhere. Not worth blocking on.

- **Content could arguably live in one file instead of six.** All six `quality-gates.md` files overlap heavily (privacy, `KOAN_ROOT`, lint). One could imagine a shared root fragment plus tiny per-skill deltas. But the feature's design is *per-skill by folder*, and the 16k cap plus "keep each skill focused" argues for the split. The current shape follows the grain of the feature — reasonable.

**Verdict:** the approach is essentially the simplest thing that works. The only genuinely optional cleanup is the `project_path or None` normalization in `fix_diagnose.py`, and that's cosmetic. The design correctly leans on machinery that already existed — the PR's real insight was that the feature was *already built* and just needed three small unblocking changes to come alive on this repo.

PR #2468, opened 2026-07-23
feat(usage): optional authoritative session/weekly quota via OAuth usage endpoint (#2455)

### LINKED ISSUE #2455: feature: authoritative session/weekly quota via Anthropic's OAuth usage endpoint
## Summary

Today Kōan's quota estimation is a **local heuristic**: it accumulates token counts scraped from the coding CLI's JSON output and divides them by a user-configured limit to produce session (5h) and weekly (7d) percentages. This works, but it carries three inherent sources of error that could be removed for users on a Claude subscription.

This issue proposes adding an **optional authoritative quota source** that reads the real utilization directly from Anthropic, to *anchor* the existing estimator rather than replace it.

## How quota estimation works today (for context)

- `app/token_parser.py` — extracts `input/output/cache` tokens from the Claude CLI `--output-format json` result after each run.
- `app/usage_estimator.py` — accumulates into `instance/usage_state.json`; models the windows in code (`DEFAULT_SESSION_LIMIT = 500_000` per 5h, `DEFAULT_WEEKLY_LIMIT = 5_000_000` per 7d; session reset on elapsed time, weekly reset on a Monday-00:00 boundary); computes `session_pct` / `weekly_pct`.
- `app/usage_tracker.py` — turns the `%` into an operating mode (`wait` / `review` / `implement` / `deep`) with a 10% safety margin.
- `app/burn_rate.py` — rolling burn-rate → time-to-exhaustion projection.
- `app/quota_handler.py` — reactive: regex-matches quota-exhaustion strings in CLI output and pauses until the parsed reset time.

The three error sources this creates:
1. **Guessed token limits** — `session_token_limit` / `weekly_token_limit` are approximations of what a plan actually allows.
2. **Manual tier calibration** — there is no Pro/Max detection; the operator hand-tunes the limits against Claude's own `/usage` and corrects drift with `/quota <N>`.
3. **Modeled reset windows** — the 5h/weekly resets are computed from wall-clock assumptions, not the real reset timestamps.

## Proposed improvement

When the user authenticates Claude Code via an OAuth subscription (Pro/Max), the CLI stores an OAuth token locally (`~/.claude/.credentials.json`, or the OS keychain under service `Claude Code-credentials`, field `claudeAiOauth`). That token can be used to query Anthropic's first-party usage endpoint:

    GET https://api.anthropic.com/api/oauth/usage
    Authorization: Bearer <cli oauth access token>
    anthropic-beta: oauth-2025-04-20

The response returns real utilization windows and reset times — `five_hour`, `seven_day`, `seven_day_opus`, `seven_day_sonnet`, and per-model scoped weekly buckets (each with a `percent` and a `resets_at`). This is exactly the data `usage_estimator` is currently approximating.

### Design: augment, don't replace

I'd suggest wiring this in as an optional authoritative source feeding the same `session_pct` / `weekly_pct` that `usage_tracker.decide_mode()` already consumes:

- **Poll periodically** (not every run — respect 429s / `retry_after`) to anchor `session_pct`, `weekly_pct`, and the real reset timestamps.
- **Keep the local token counter running between polls** as the interpolator, so decisions stay responsive within the tight autonomous loop.
- **Fall back to the current heuristic** whenever the authoritative source is unavailable — this is essential because it does **not** exist for:
  - users on `ANTHROPIC_API_KEY` (no OAuth token),
  - non-Claude providers (Codex / Copilot / Ollama).
- **Keep `quota_handler.py` as the last-resort safety net.**

Behind a config flag, e.g.:

    usage:
      authoritative_source: auto   # auto | oauth_usage | off

### Caveats worth noting

- The endpoint / `oauth-2025-04-20` beta header are **not a documented, stable API** — treat availability as best-effort and always degrade gracefully to the heuristic.
- The figures are **account-wide** (they reflect all Claude usage, not just Kōan's), so per-run attribution should stay on the local counter; the authoritative source is for the "how full is my plan right now" gate.

### Rough scope

A small `oauth_usage` client module + a source-selection shim in `usage_tracker` (prefer authoritative `%`/reset when fresh, else fall back to the estimator). No changes needed to the mode logic, burn-rate, or reactive handler.

## Why this matters

Because Kōan can be setup to use someone's account own OAuth token, **this feature would let Kōan have an accurate estimation of session usage whenever it is used alongside Claude with the same token** — reading the true 5h/weekly consumption straight from Anthropic instead of inferring it from a hand-calibrated local guess, while still falling back cleanly to the existing heuristic for API-key users and non-Claude providers. Furthermore, this can have an accurate estimate when Anthropic performs session resets for their users.

MaxBohm-YourArt:
Closes #2455.

## Summary

Adds an **optional** authoritative session/weekly quota source that reads real account utilization from Anthropic's OAuth usage endpoint and **anchors** the existing local heuristic estimator rather than replacing it. Fully backward-compatible and behind a config flag defaulting to `auto`.

**New modules**
- `koan/app/oauth_usage.py` — OAuth usage client. Reads the Claude CLI's OAuth **access** token (`~/.claude/.credentials.json`, then the macOS keychain service `Claude Code-credentials`; field `claudeAiOauth.accessToken`) and GETs `https://api.anthropic.com/api/oauth/usage` with the `anthropic-beta: oauth-2025-04-20` header + `claude-code/<version>` UA. Maps `five_hour`→session and `seven_day`→weekly (per-model buckets preserved). Honors **429** `Retry-After`/backoff and re-reads the CLI-rotated access token **once** on **401**. Never spends or rotates the refresh token; every failure returns `None`.
- `koan/app/authoritative_usage.py` — source-selection shim. On a successful poll it stores an **anchor** (`instance/.oauth-usage.json`): account-wide %/real reset timestamps + the local token-counter values at poll time. Between polls the **local counter interpolates** on top of the anchor (`anchor_pct + Δtokens/limit·100`), so per-run attribution and burn-rate stay on the local counter (OAuth figures are account-wide). Polls periodically (`authoritative_poll_seconds`, default 300s), not per-run.

**Integration**
- `usage_estimator._write_usage_md` prefers the authoritative %/reset when fresh, keeping percentages **integer** (the tracker parser matches `(\d+)%`) and recording the chosen source in an HTML comment.
- `decide_mode`, `burn_rate`, and `quota_handler` are **untouched** — `quota_handler` remains the last-resort reactive safety net.

**Mandatory graceful degradation** — falls back to the heuristic for: `authoritative_source: off`, API-key users (no OAuth token), non-Claude / no-quota providers (`has_api_quota()` false: Codex/Copilot/Ollama), any HTTP error, or an anchor older than `authoritative_max_staleness_seconds` (default 900s) / past its window reset. Session and weekly windows degrade independently.

**Config** — `usage.authoritative_source: auto | oauth_usage | off` (default `auto`) plus optional `authoritative_poll_seconds` / `authoritative_max_staleness_seconds` in `instance.example/config.yaml`.

**Caveats respected in code/comments** — the endpoint and the `oauth-2025-04-20` beta header are undocumented/unstable; availability is best-effort.

## Testing

- `make lint` — clean.
- New tests (mock the HTTP client + credential reads; no network/keychain): `koan/tests/test_oauth_usage.py`, `koan/tests/test_authoritative_usage.py`, plus estimator-shim tests in `koan/tests/test_usage_estimator.py`. Cover: authoritative path chosen when fresh, fallback on missing token / non-Claude provider / HTTP error / staleness / window reset, **401 re-read**, **429 backoff**, interpolation clamping, and all three config flag values.
- Regression: `usage_estimator`, `usage_tracker`, `burn_rate`, `quota_handler`, `provider_quota`, `iteration_manager`, `mission_runner` all pass; full collection clean (19,481 tests).

Docs/specs updated contract-first: `specs/components/agent-loop.md` § "Usage source selection", `docs/architecture/daemon.md`, `instance.example/config.yaml`, `koan/app/CLAUDE.md` instance-state list.

## Declarations

- [x] **Architectural change** — this PR modifies a durable design contract (`specs/components/**` or `specs/skills/**`). The new architecture needs review before approval. Rationale: adds a new usage source-selection contract to `specs/components/agent-loop.md` (authoritative OAuth anchor + local interpolation), landed contract-first.


MaxBohm-YourArt:
Thanks for the review — pushed `a7b362a6` addressing the blocking items and the silent-failure notes.

**🟡 Important #1 — failed polls stalling the loop.** Fixed. The anchor now carries a `last_attempt_at` (persisted in `.oauth-usage.json`) alongside the last-success `polled_at`. `maybe_poll` gates the fetch on `attempt_at`, so a failed/rate-limited poll is **negatively cached** for `authoritative_poll_seconds` instead of re-triggering the synchronous fetch on every mission. First-attempt failures with no prior anchor persist a `polled_at=0` marker that `resolve()` treats as stale → heuristic. New tests: `test_maybe_poll_failed_poll_is_negatively_cached`, `test_maybe_poll_first_attempt_failure_negatively_cached`, `test_negatively_cached_anchor_resolves_to_heuristic`.

**🟡 Important #2 — real `claude --version` subprocess in tests.** Fixed. Added an autouse fixture in `test_oauth_usage.py` that stubs `_claude_cli_version`, so the `_http_get_*` tests never probe the CLI. Suite is hermetic regardless of whether `claude` is on PATH.

**Silent-failure (MEDIUM) — catch-all in `usage_estimator`.** Fixed. Now logs at `WARNING` with `exc_info=True` via the module logger, so a permanently-broken `resolve()` is distinguishable from normal heuristic fallback in daemon logs (still never raises — usage.md writes must not fail).

**Silent-failure (MEDIUM) — `_state_int` zero fallback.** A *present but unparseable* counter is now debug-logged before falling back to 0 (counter corrupt = under-report = the dangerous direction, so it's observable).

**🟢 Suggestion #2 — utilization scale (0–100 vs 0–1).** `_parse_window` now debug-logs a sub-1.0 utilization value so a scale mismatch on the undocumented endpoint is diagnosable in the field. Kept the clamp-only behavior (no auto-scaling — a genuine 0.5% must not be silently ×100'd).

**🟢 Suggestion #1 — `auto` default.** Kept `authoritative_source: auto` per issue #2455's explicit spec ("default auto"). Degradation is graceful and the credential read is best-effort; happy to flip to `off`-by-default if you'd prefer the more conservative stance.

`make lint` clean; 915 tests pass across the OAuth/authoritative/estimator/tracker/burn-rate/quota/iteration/mission suites.

MaxBohm-YourArt:
Pushed `59160d0f` implementing **MEDIUM #2** (unparseable token counter).

`_read_counter` now returns `None` for a present-but-unparseable counter (WARNING-logged) vs `0` for absent/empty, and `_resolve_window` treats a `None` counter as un-interpolatable — falling back to the heuristic for **that window only** rather than interpolating from `0` (which under-reports usage, the dangerous direction). The anchor baseline at poll time keeps a `0` coercion, since a later delta against a good counter over-reports (the safe direction). New tests cover valid/absent/empty/corrupt counters and the per-window fallback.

On the other two, I'm intentionally **not** taking the proposed changes:

- **MEDIUM #1 (sub-1.0 utilization → return None / warn).** Real accounts sit below 1% routinely (start of every session), so returning `None` on sub-1 would discard legitimate low-usage data (a false negative on normal input) and "warn once" would be constant noise. The 0–100 vs 0–1 question can only be settled with an actual response sample from the undocumented endpoint, not a heuristic guard — so I've kept the diagnostic debug log and left the value untouched.
- **MEDIUM #3 (persist failure defeats negative cache).** Real but narrow: it only bites if `instance/` is unwritable, at which point the daemon is already broken (it writes `missions.db`/`usage.md`/etc. there every iteration). The in-memory-timestamp fix trades module-global mutable state for a scenario that effectively can't occur in a running daemon. The higher-leverage guard for the underlying "blocking fetch on the hot path" concern is bounding the 429 backoff budget (the 🟢 suggestion) — happy to do that as a small follow-up if you'd like, but it's non-blocking.

`make lint` clean; affected suites green (164 tests across oauth/authoritative/estimator).

atoomic:
@Koan-Bot explain

atoomic (approved):

sukria:
Thanks @MaxBohm-YourArt for the contribution! Very nice finding.

PR #2480, opened 2026-07-28
feat(context): one deduped steering summary before each Claude call

### LINKED ISSUE #2469: Better logging about local KOAN.md files before Claude Code is invoked, remove duplicate logging messages.
Linked to #2463 : KOAN.md notice is displayed many times:

````
==> logs/run.log <==
[koan] [memory] FTS5 selected 20/32 learnings for marketplace — task='## Original Issue #378: Add translation test\n\nAdd a new tran'
[context] Detected KOAN.md, loaded 2997 chars (~ 856 tokens)
[context] Detected KOAN.md, loaded 2997 chars (~ 856 tokens)

⚠️  Claude hit the max turns limit (3). The output may be incomplete.
   This call uses a hardcoded limit and is not configurable.

[plan_runner] Review round 1: APPROVED
[context] Detected KOAN.md, loaded 2997 chars (~ 856 tokens)
Assumptions check returned unparseable output: "All the plan's load-bearing factual claims check out against the actual repository state."
Assumptions audit skipped (reviewer error): assumptions check produced unparseable output — manual review needed

[plan] Starting plan runner
````

Would be better to un that, would also be super handy to output a summary of all local .md files loaded before CLI Claude is invoked?

### PR DESCRIPTION
**Pull request by sukria-koan0**
## Summary

`/plan` and other multi-CLI-call skills rebuild their prompt several times per mission,
and every rebuild re-announced the same steering file — so `make logs` showed
`Detected KOAN.md` two or three times for a single mission. This replaces the per-file
`Detected …` lines with an **accumulate-then-summarize** model: each prompt build
silently records the steering files it folds in, then emits **one** consolidated summary
right before the CLI is invoked. Repeats within a mission are suppressed; each new
mission re-announces.

Before:

```
[context] Detected KOAN.md, loaded 1240 chars (~ 354 tokens)
[context] Detected KOAN.md, loaded 1240 chars (~ 354 tokens)
[context] Detected .koan/skills/plan, loaded 380 chars (~ 108 tokens)
```

After:

```
[context] Steering loaded before Claude (3 file(s), ~ 3004 tokens total):
[context]   • CLAUDE.md (auto-loaded by CLI): 8900 chars (~ 2542 tokens)
[context]   • KOAN.md: 1240 chars (~ 354 tokens)
[context]   • .koan/skills/plan: 380 chars (~ 108 tokens)
```

Closes https://github.com/Anantys-oss/koan/issues/2469

## Changes

- `project_koan.py`: `log_context_load` replaced by `record_context_load` (silent
  accumulate, deduped by label), `flush_context_summary` (emits one block, suppresses a
  summary identical to the previous one, clears the batch) and `reset_context_load_log`
  (per-mission reset). Every path stays best-effort and never raises.
- `prompt_builder.py` / `prompts.py`: the four call-sites now record; `build_agent_prompt`,
  `build_agent_prompt_parts` and `load_skill_prompt` flush once at the end of the build,
  immediately before the CLI call.
- **Gap closed**: `build_agent_prompt_parts` (the split builder the agent loop actually
  uses) never ran `_log_claude_md_detected`, so `CLAUDE.md` was invisible on that path. It
  is now recorded there too — the summary really covers *all* local `.md` files.
- `mission_executor._run_iteration` calls `reset_context_load_log()` so a fresh mission
  re-announces byte-identical steering context.
- Specs updated **contract-first** (`specs/components/agent-loop.md`,
  `specs/components/skills.md`), then the code made to conform. `docs/users/koan-md.md`
  shows the new block.

Out of scope per the issue thread: thread-local batching for parallel sessions, and
wrapping contemplative sessions in the per-mission reset.

## Test plan

- New unit matrix in `koan/tests/test_project_koan.py`: record is silent → flush emits;
  empty batch is a no-op; per-label dedup within a build; consecutive identical summary
  suppressed; a changed steering set re-emits; reset lets the next mission re-emit; a
  broken token estimator never raises.
- `test_prompts.py` / `test_prompt_builder.py` updated to assert the summary through the
  real flush points.
- Autouse conftest fixture resets the module-global batch so the dedup signature cannot
  bleed between tests.
- `KOAN_ROOT=/tmp/test-koan .venv/bin/pytest koan/tests/test_project_koan.py
  koan/tests/test_prompts.py koan/tests/test_prompt_builder.py` → 319 passed.
- `make lint` → clean. `scripts/wiki_check.py` → no gaps.
- One pre-existing failure on `main` is unrelated and untouched
  (`test_messaging_level.py::test_notify_outcome_always_logs_and_sends`).

## Declarations

- [x] **Architectural change** — this PR modifies a durable design contract
  (`specs/components/agent-loop.md`, `specs/components/skills.md`). The new architecture
  needs review before approval. Rationale: the steering-visibility contract changes from
  "announce each file at read time" to "record silently, flush one deduped summary per
  prompt build, reset per mission".

---

### DIFF
--- a/docs/users/koan-md.md
+++ b/docs/users/koan-md.md
@@ -4,7 +4,7 @@ title: "KOAN.md — koan-only project instructions"
 description: "Documents the optional project-root KOAN.md file and the .koan/ directory (a second .koan/KOAN.md, per-skill .koan/skills/<skill>/*.md hooks, and a structured .koan/config.yaml with review.always_check): koan-only steering injected into the autonomous agent's system prompt but never loaded by interactive Claude Code sessions, with precedence rules, the 16k-char cap, and this repo's dogfood layout."
 tags: [users]
 created: 2026-07-09
-updated: 2026-07-26
+updated: 2026-07-28
 ---
 
 # KOAN.md — koan-only project instructions
@@ -80,19 +80,26 @@ the mission text instead.
 
 ### Seeing what got loaded (`make logs`)
 
-Every steering file koan folds into a prompt is announced on the run log, so you
-can confirm the context is actually in play. Watch `make logs` for lines like:
+Every steering file koan folds into a prompt is announced on the run log as a
+single summary before Claude Code runs, so you can confirm the context is
+actually in play. Watch `make logs` for a block like:
 
 ```
-[context] Detected KOAN.md, loaded 1240 chars (~ 354 tokens)
-[context] Detected .koan/skills/plan, loaded 380 chars (~ 108 tokens)
-[context] Detected CLAUDE.md (auto-loaded by CLI), loaded 8900 chars (~ 2542 tokens)
+[context] Steering loaded before Claude (3 file(s), ~ 3004 tokens total):
+[context]   • CLAUDE.md (auto-loaded by CLI): 8900 chars (~ 2542 tokens)
+[context]   • KOAN.md: 1240 chars (~ 354 tokens)
+[context]   • .koan/skills/plan: 380 chars (~ 108 tokens)
 ```
 
-`KOAN.md` and `.koan/skills/<skill>` lines mean koan injected that content;
-the `CLAUDE.md` line is **detection-only** — Claude Code loads `CLAUDE.md` from
+`KOAN.md` and `.koan/skills/<skill>` entries mean koan injected that content;
+the `CLAUDE.md` entry is **detection-only** — Claude Code loads `CLAUDE.md` from
 the working directory itself, koan just reports its size so you can see the full
 steering context at a glance. Token counts are a `chars/3.5` estimate.
+
+The block appears once per mission: skills that call the CLI several times (like
+`plan`) rebuild the prompt for each call, and a repeat of the same steering set
+is suppressed rather than re-printed. A new mission always re-announces, even
+when the files have not changed.
 
 ## The `.koan/config.yaml` file
 
--- a/koan/app/mission_executor.py
+++ b/koan/app/mission_executor.py
@@ -701,6 +701,11 @@ def _run_iteration(
 
     run_num = count + 1
 
+    # Fresh steering-summary scope per mission: a new mission re-announces its
+    # loaded .md context even when byte-identical to the previous mission's.
+    from app.project_koan import reset_context_load_log
+    reset_context_load_log()
+
     # --- Parallel session reap (Phase 1) ---
     # Poll any active parallel sessions and process completions before planning
     # the next iteration.  Skipped when max_parallel_sessions == 1 (default) so
--- a/koan/app/project_koan.py
+++ b/koan/app/project_koan.py
@@ -32,29 +32,87 @@ _MAX_HOOK_CMD_LEN = 1000
 _HOOK_PHASES = {"pre": "pre_hooks", "post": "post_hooks"}
 
 
-def log_context_load(label: str, content: str) -> None:
-    """Announce a steering file koan just loaded into a prompt, for ``make logs``.
+# --- Steering-context load accumulation (per prompt build) -------------------
+# Each prompt build records the steering files it folds in via
+# record_context_load(); flush_context_summary() emits ONE consolidated block and
+# clears the batch. A summary identical to the previous one is suppressed, so a
+# multi-CLI-call skill (plan builds 3 prompts per mission) announces its steering
+# context once, not once per CLI call. reset_context_load_log() clears the dedup
+# memory per mission so a fresh mission re-announces byte-identical context.
+_pending_loads: list = []   # [(label, chars, tokens), ...] for the current build
+_last_summary_sig = None    # signature of the last-emitted summary
 
-    Emits ``Detected <label>, loaded N chars (~ M tokens)`` on **stderr** so it
-    lands in ``logs/run.log`` (visible via ``make logs``) without ever
-    corrupting the JSON some skill runners write to stdout. The ``logging``
-    module has no stdout/stderr handler wired in the run loop, so ``logger.info``
-    alone would be invisible there — hence the direct ``print``.
 
-    Best-effort: a broken stream (or a missing ``estimate_tokens``) must never
-    break prompt assembly, so every failure is swallowed — logged at debug so it
-    stays visible without ever raising.
-    """
+def _estimate_tokens(content: str) -> int:
     try:
         from app.diff_compressor import estimate_tokens
-        print(
-            f"[context] Detected {label}, loaded {len(content)} chars "
-            f"(~ {estimate_tokens(content)} tokens)",
-            file=sys.stderr,
-            flush=True,
-        )
+        return estimate_tokens(content)
+    except Exception:
+        return 0
+
+
+def record_context_load(label: str, content: str) -> None:
+    """Silently record a steering file folded into the prompt about to be sent.
+
+    Accumulates ``(label, chars, tokens)`` into the current build's batch, deduped
+    by ``label`` (a re-read of the same file updates in place). The consolidated
+    line is emitted later by :func:`flush_context_summary`. Best-effort: never
+    raises — a broken estimator must not break prompt assembly.
+    """
+    try:
+        entry = (label, len(content), _estimate_tokens(content))
+        for i, (lbl, _chars, _tokens) in enumerate(_pending_loads):
+            if lbl == label:
+                _pending_loads[i] = entry
+                return
+        _pending_loads.append(entry)
     except Exception as e:
-        logger.debug("log_context_load failed for %s: %s", label, e)
+        logger.debug("record_context_load failed for %s: %s", label, e)
+
+
+def flush_context_summary() -> None:
+    """Emit one consolidated steering-context summary for the current build.
+
+    Prints on **stderr** (→ ``logs/run.log``, visible via ``make logs``) a block
+    listing every steering file recorded since the last flush, with per-file and
+    total token estimates, then clears the batch. stderr keeps the block away from
+    the JSON some skill runners write to stdout; ``logging`` has no handler wired
+    in the run loop, so a direct ``print`` is the only thing operators would see.
+    Suppresses a summary identical to the previous one (same files + sizes) so
+    repeated prompt builds within a single mission announce once. No-op on an
+    empty batch. Best-effort: never raises.
+    """
+    global _last_summary_sig
+    try:
+        if not _pending_loads:
+            return
+        sig = tuple(_pending_loads)
+        batch = _pending_loads.copy()
+        _pending_loads.clear()
+        if sig == _last_summary_sig:
+            return
+        _last_summary_sig = sig
+        total = sum(tokens for _lbl, _chars, tokens in batch)
+        lines = [
+            f"[context] Steering loaded before Claude "
+            f"({len(batch)} file(s), ~ {total} tokens total):"
+        ]
+        for lbl, chars, tokens in batch:
+            lines.append(f"[context]   • {lbl}: {chars} chars (~ {tokens} tokens)")
+        print("\n".join(lines), file=sys.stderr, flush=True)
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

**Comment by atoomic** [comment]
@sukria-koan0 fix

**Comment by sukria-koan0** [comment]

Branch `koan0/implement-2469` was rebased onto `main` and review feedback was applied.

### Changes applied

- Lint clean (`make lint` itself failed on offline `pip install ruff`; ran ruff directly).
- `project_koan._estimate_tokens`: log the swallowed exception at debug before returning 0, restoring the trace the old `log_context_load` emitted (silent-failure finding).
- `project_koan`: comment at `_pending_loads`/`_last_summary_sig` naming the single-builder assumption, the log-only blast radius, and `threading.local()` as the follow-up (suggestion #1).
- `test_prompt_builder.py`: new `test_steering_summary_lists_claude_md_then_koan_md` — real `build_agent_prompt_parts()` against a tmp project with `CLAUDE.md` + `KOAN.md`, asserting both appear in the flushed summary in that order (suggestion #2).

### Stats

```
11 files changed, 280 insertions(+), 65 deletions(-)
```

<details>

- Already-solved check: negative (confidence=high, reasoning=main still has per-file `log_context_load` emitting `Detected <label>` lines with no accumulate/flus)
- Rebased `koan0/implement-2469` onto `origin/main`
- Applied review feedback
- Pre-push CI check: previous run #30336480614 failed
- Pre-push CI fix: no changes needed or Claude found nothing to fix
- Force-pushed `koan0/implement-2469` to origin

</details>

### CI status

---
_Automated by [Skuggi](https://skuggi.ai)_

**Comment by Master-Koan** [comment]
## 💡 PR Explanation

## 1. The Problem

### What the code did before

Kōan injects "steering" files into the prompt it sends to Claude Code:

- `KOAN.md` (project-root, koan-only instructions) — **injected** by koan
- `.koan/skills/<skill>/*.md` (per-skill hooks) — **injected** by koan
- `CLAUDE.md` — **not** injected; the CLI auto-loads it from the working directory. Koan only reports its size so you can see the full picture.

Every time one of those files was read, `project_koan.log_context_load()` printed one line to stderr (which lands in `logs/run.log`, so `make logs` shows it):

```
[context] Detected KOAN.md, loaded 1240 chars (~ 354 tokens)
```

That's a "print at read time" design: **one line per read**, no memory of what was already printed.

### Why it broke

A prompt isn't built once per mission. Some skills call the CLI several times, and each call rebuilds the prompt from scratch — so each rebuild re-reads `KOAN.md` and re-announces it.

**Concrete example** — run `/plan` on a project that has a `KOAN.md` and a `.koan/skills/plan/` hook. `make logs` shows:

```
[context] Detected KOAN.md, loaded 1240 chars (~ 354 tokens)
[context] Detected KOAN.md, loaded 1240 chars (~ 354 tokens)
[context] Detected .koan/skills/plan, loaded 380 chars (~ 108 tokens)
```

Same file, twice, for one mission. Nothing was wrong with the prompt — only the log lied about how many times context was loaded, and the noise buried the real signal.

**A second, worse bug (silent):** there are two prompt builders in `koan/app/prompt_builder.py`:

- `build_agent_prompt()` — the single-string builder, used by the CLI entrypoint and most tests
- `build_agent_prompt_parts()` — the **split** system/user builder, which is what the agent loop actually calls (`mission_executor.py:1273`)

Only the first one called `_log_claude_md_detected()`. So on the path that runs in production, `CLAUDE.md` was **never** reported. An operator reading `make logs` would conclude their `CLAUDE.md` wasn't in play — while the CLI was loading it the whole time. Wrong information is worse than no information.

**Impact:** log noise plus a misleading picture of the agent's context budget. No behavior change for Claude — this is purely observability.

---

## 2. The Solution

The core move: **stop printing at read time; accumulate, then print once.**

### `koan/app/project_koan.py` — the accumulator

`log_context_load()` is replaced by three functions and two module-level globals:

- **`record_context_load(label, content)`** — silent. Appends `(label, chars, tokens)` to `_pending_loads`. If the same `label` is already in the batch, it **overwrites in place** instead of appending, so a re-read within one build can't double-count.
- **`flush_context_summary()`** — prints one consolidated block to stderr, then clears the batch. If the batch is byte-for-byte the same set it printed last time (`_last_summary_sig`), it stays silent.
- **`reset_context_load_log()`** — clears both the batch and the dedup signature.
- **`_estimate_tokens()`** — extracted helper; on failure it logs at `debug` and returns `0` instead of blowing up.

Every path is wrapped in `try/except` — the pre-existing contract that logging **never** breaks prompt assembly is preserved.

### `koan/app/prompt_builder.py` and `koan/app/prompts.py` — record + flush points

- The four call sites (`_get_koan_md_section`, `_log_claude_md_detected`, `_maybe_append_project_skill_instructions`, `_maybe_append_general_koan_md`) now call `record_context_load` instead of `log_context_load`. Mechanical rename.
- Three builders flush once at the end: `build_agent_prompt()`, `build_agent_prompt_parts()`, `prompts.load_skill_prompt()`.
- **The real bug fix:** `build_agent_prompt_parts()` now calls `_log_claude_md_detected(host_project_path or project_path)` — placed *before* the `KOAN.md` read so the summary ordering matches the other builder.

### `koan/app/mission_executor.py` — per-mission reset

`_run_iteration()` calls `reset_context_load_log()` at the top. Without this, the dedup signature would survive across missions and mission #2 would never announce its (identical) steering set. Note `_run_iteration` is one pass of the main loop — at most one mission — so "per mission" is accurate.

### Tests

- `koan/tests/test_project_koan.py` gets a behavior matrix: recording is silent, flush emits, empty batch is a no-op, same label dedups, identical consecutive summary is suppressed, a *changed* set re-emits, reset re-enables emission, broken estimator never raises.
- `koan/tests/test_prompt_builder.py` adds `test_steering_summary_lists_claude_md_then_koan_md` — a builder-level test that actually exercises `build_agent_prompt_parts` and asserts `CLAUDE.md` appears before `KOAN.md`. **This is the test that guards the headline gap**; without it the fix would be one line nobody notices regressing.
- `koan/tests/conftest.py` gets an autouse fixture resetting the globals around every test — module-level mutable state leaks across tests otherwise.

### Before / after

```
# Before — one mission, /plan
[context] Detected KOAN.md, loaded 1240 chars (~ 354 tokens)
[context] Detected KOAN.md, loaded 1240 chars (~ 354 tokens)
[context] Detected .koan/skills/plan, loaded 380 chars (~ 108 tokens)
# (CLAUDE.md: never shown on this path)

# After
[context] Steering loaded before Claude (3 file(s), ~ 3004 tokens total):
[context]   • CLAUDE.md (auto-loaded by CLI): 8900 chars (~ 2542 tokens)
[context]   • KOAN.md: 1240 chars (~ 354 tokens)
[context]   • .koan/skills/plan: 380 chars (~ 108 tokens)
```

The `~ 3004 tokens total` rollup is new and genuinely useful — that's your steering overhead per call, at a glance.

---

## 3. How It Works

### End-to-end: one `/plan` mission

```mermaid
flowchart TD
    A["_run_iteration()<br/>reset_context_load_log()"] --> B["build_agent_prompt_parts()"]
    B --> C["_log_claude_md_detected()<br/>record: CLAUDE.md"]
    C --> D["_get_koan_md_section()<br/>record: KOAN.md"]
    D --> E["flush_context_summary()<br/>PRINTS block, saves signature"]
    E --> F["CLI call #1"]
    F --> G["Prompt rebuilt for call #2<br/>records the same 2 files"]
    G --> H["flush: signature matches<br/>SILENT, batch cleared"]
    H --> I["CLI call #2"]
    I --> J["next mission:<br/>reset clears signature"]
    J --> K["flush PRINTS again"]
```

1. **Reset.** A new loop iteration wipes the batch and the "last printed" signature.
2. **Record.** Each steering read appends silently. `_pending_loads` holds tuples; dedup is by `label`.
3. **Flush.** At the end of the build: if the batch is empty → return. Otherwise compute the signature, copy the batch, clear it, compare. Match → silent return. No match → save signature, sum tokens, print the block.
4. **Repeat builds** in the same mission carry the same files at the same sizes → identical signature → silent.
5. **New mission** → reset nulls the signature → the block prints again even if the files never changed.

### Edge cases handled

- **Empty batch** (project has no `KOAN.md`, no `.koan/`) → nothing printed at all, same as before.
- **Broken token estimator** → `_estimate_tokens` returns `0`, prompt assembly unaffected.
- **Same file read twice in one build** → single entry, in-place update.
- **Changed steering set** mid-mission (e.g. `plan`'s later call adds a skill hook) → signature differs → re-emits.
- **stderr, not stdout** — deliberate: some skill runners write JSON to stdout, and a log line there would corrupt the parse. This constraint is unchanged from the original code and is why a raw `print` is used instead of `logger.info` (the run loop wires no stream handler for `logging`).

### Documented non-goals

The PR explicitly declines two things: thread-local batching for parallel sessions (`max_parallel_sessions > 1` would garble the *log*, never the prompt), and wrapping contemplative sessions in the reset. Both are stated in the module comment and the PR body. Naming a limitation in a comment beats pretending it doesn't exist.

---

## 4. Could It Be Simpler?

**Yes — meaningfully so. Two of the three moving parts exist only because state was made global.**

### The two changes worth the most, and their cost

The genuinely valuable fix is **two lines** in `build_agent_prompt_parts()` (record `CLAUDE.md`) plus the test that guards it. That's the bug an operator would actually be misled by. It could have shipped by itself.

The duplicate-line fix is the rest of the PR. A minimal version keeps `log_context_load()` and adds a seen-set:

```python
_seen: set = set()

def log_context_load(label, content):
    key = (label, len(content))
    if key in _seen:
        return
    _seen.add(key)
    print(...)
```

Roughly five lines, no flush points, no ordering question, no leaked-batch failure mode. **Trade-off:** you lose the `~ T tokens total` rollup and the grouped block. That rollup is real value — it's the number that tells you how much of the context window steering eats before the mission text even starts. So the accumulate-then-summarize model is defensible; it just isn't free.

### The design smell: a module global where a local would do

`_pending_loads` and `_last_summary_sig` are process-wide mutable state. That single choice causes:

- `reset_context_load_log()` to exist at all
- the autouse fixture in `koan/tests/conftest.py`
- the "NOT thread-safe" caveat in the module comment
- **the leak the review flagged** — `flush_context_summary()` is called at the *end* of the builder, not in a `finally`. If anything between the first `record_context_load` and the flush raises (a missing `koan-md` template, say), the entries survive and the *next* build's summary reports files that aren't in the current prompt.

The alternative: builders collect entries in a **local list** and call `emit_context_summary(entries)` once. The builders already know every record site, since they call them. That deletes the global, the reset, the fixture, and the thread-safety footnote in one stroke. `prompts.load_skill_prompt` is the awkward case — recording happens inside two private helpers (`_maybe_append_project_skill_instructions`, `_maybe_append_general_koan_md`) that currently return only a string — but those are private to the module and could return `(prompt, entry)`. That's a small local refactor, not a redesign.

### Two smaller issues, both flagged by the automated review and both still present in the diff

- **Signature committed before the print.** `flush_context_summary()` sets `_last_summary_sig = sig` *before* calling `print`. If the print raises, the exception is swallowed — the summary is lost **and** the signature is recorded, so the next identical build is suppressed too. The block silently never appears. Moving the assignment after the successful `print` fixes it.
- **Dedup is "consecutive only."** The check is against one previous signature, not a set of seen ones. A build sequence A → B → A within one mission prints A twice. Probably fine in practice; worth knowing it's not full dedup.

### Minor redundancy

`koan/tests/test_project_koan.py` adds its own autouse `_reset_context_log` fixture even though `conftest.py` now has an identical autouse fixture covering every test. One of the two is dead weight.

### Verdict

The approach is sound and the observability win is real. But the PR is roughly **twice the machinery it needs** — the global-state choice is what pulls in the reset function, the conftest fixture, the thread-safety caveat, and the leak-on-exception path. A local-batch variant gets the same output with a strictly smaller failure surface.

atoomic:
@sukria-koan0 rebase

